PR #25, opened 2022-08-02
feat: exchange pricefeed service

enigbe:
## What this PR does?
- Fetches price data from an OKEX channel and publish the said data to pubsub channel

### DIFF
--- a/okex-price/src/pricefeed/feeder.rs
+++ b/okex-price/src/pricefeed/feeder.rs
@@ -1,0 +1,46 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ChannelArgs {
+    pub channel: String,
+    pub inst_id: String,
+}
+
+#[derive(serde::Serialize)]
+pub struct Subscribe {
+    op: String,
+    args: Vec<ChannelArgs>,
+}
+
+impl Subscribe {
+    pub fn new() -> Self {
+        Self {
+            op: "subscribe".to_string(),
+            args: vec![ChannelArgs {
+                channel: "tickers".to_string(),
+                inst_id: "BTC-USD-SWAP".to_string(),
+            }],
+        }
+    }
+}
+
+impl Default for Subscribe {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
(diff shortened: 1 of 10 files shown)


okex-price/src/pricefeed/feeder.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Clone, Serialize, Deserialize, Debug)]
+pub struct OkexPriceTick {
+    arg: ChannelData,
+    data: Vec<TickersChannelData>,
+}
+
+payload! { OkexPriceTick, "okex.price" }
+
+#[derive(Debug)]
+pub struct PriceFeed {
```
bodymindarts, line 71: I guess we don't need the PriceFeed struct as we cannot re-use the receiver.

I guess we can have a top level:
src/price-feed/mod.rs
```
pub fn subscribe_btc_usd_swap()
```
or something like that

okex-price/src/pricefeed/feeder.rs
```diff
@@ -0,0 +1,107 @@
...
+        };
+        let (ws_stream, _) = connect_async(config.url).await?;
+        let (sender, receiver) = ws_stream.split();
+
+        Ok(Self { sender, receiver })
+    }
+
+    pub async fn subscribe_btcusdswap(
+        &mut self,
+    ) -> Result<Pin<Box<dyn Stream<Item = Envelope<OkexPriceTick>> + Send>>, PriceFeedError> {
```
bodymindarts, line 89: `Envelope` is for `pubsub` serialisation - not what re receive from okex.

okex-price/tests/pricefeed.rs
```diff
@@ -0,0 +1,34 @@
+mod tests {
```
bodymindarts, line 1: no need for `mod tests` here. That is only for unit tests embedded in the files under `src` anything under `tests` is already a test.

okex-price/src/pricefeed/feeder.rs
enigbe: I thought it would be a good idea to have metadata and payload type on the data received from okex but I will remove it and deserialize `OkexPriceTick` without the wrapper.

okex-price/src/pricefeed/feeder.rs
enigbe: I have implemented the `subscribe_btc_usd_swap()` function as recommended.

PR #34, opened 2022-08-04
feat: cli can run price server

bodymindarts:


stablesats.yml
```diff
@@ -1,2 +1,8 @@
 pubsub:
   host: localhost
+
+## defaults:
```
tvpeter, line 4: Wondering since this file will serve as the default for configs, is there need to comment the below lines? or can the CI runner be able to pick them up that way?

enigbe (approved):

stablesats.yml
bodymindarts: That’s file servers 2 purposes:

- a config that the Cli will pick up if running in the root of the repo (for example when your testing the Billy artifact locally)
- a kind of Schema documentation for ppl finding this repo wondering what can be configured and how.


The actual defaults are defined in code

PR #35, opened 2022-08-04
test: add mid price fn to ExchangePriceCacheInner

tvpeter:
- add mid-price for bid and ask price
- add unit test for bid, ask, and mid-price

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@
 - More price-server scafolding
 - Return pined stream from subscribe
 - Clippy
-- Commit Cargo.lock as we are buiding a binary
+- Commit Cargo.lock as we are building a binary
 - Missing -
 - Add check-code
 - Remove --all-features from watch commands
--- a/price-server/src/app/mod.rs
+++ b/price-server/src/app/mod.rs
@@ -8,7 +8,7 @@ pub use error::*;
 use shared::{currency::*, payload::OkexBtcUsdSwapPricePayload, pubsub::*};
 
 pub struct PriceApp {
-    price_cache: ExchangePriceCache,
+    pub price_cache: ExchangePriceCache,
 }
 
 impl PriceApp {
@@ -33,7 +33,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.ask_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.ask_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
@@ -41,7 +41,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.bid_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.bid_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
@@ -49,7 +49,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.ask_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.ask_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
@@ -57,7 +57,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.bid_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.bid_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
@@ -65,7 +65,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.ask_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.ask_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
@@ -73,7 +73,7 @@ impl PriceApp {
         &self,
         sats: impl Into<Sats>,
     ) -> Result<u64, PriceAppError> {
-        let cents = self.price_cache.bid_price_of_one_sat().await?;
+        let cents = self.price_cache.latest_tick().await?.bid_price_of_one_sat;
         Ok(u64::try_from(cents * *sats.into().amount())?)
     }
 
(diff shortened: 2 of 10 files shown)


price-server/src/server/mod.rs
```diff
@@ -135,9 +135,18 @@ impl Price for PriceService {
...
+        request: Request<GetCentsPerSatsExchangeMidRateRequest>,
     ) -> Result<Response<GetCentsPerSatsExchangeMidRateResponse>, Status> {
-        Err(Status::unimplemented(""))
+        let req = request.into_inner();
+
+        let mid_amount_in_cents = self
+            .app
+            .get_cents_per_sats_exchange_mid_rate(Sats::from_major(req.amount_in_satoshis))
+            .await?;
+        let ratio_in_cents_per_satoshis = f64::from_bits(mid_amount_in_cents);
```
bodymindarts, line 146: this line looks dangerous. I would prefer a more explicit conversion (that is unit tested).

price-server/src/exchange_price_cache.rs
```diff
@@ -95,4 +100,61 @@ impl ExchangePriceCacheInner {
         }
         Err(ExchangePriceCacheError::NoPriceAvailable)
     }
+
+    fn mid_price_of_one_sat(&self) -> Result<UsdCents, ExchangePriceCacheError> {
```
bodymindarts, line 104: What you've done here is correct - following the existing pattern...

however I think we should cleanup the `ExchangePriceCache`.

It should just expose the `BtcSatTick` directly:
```
impl ExchangePriceCache {
  pub fn get_latest_tick()
}
```

Then in the us case you can call:
```
cache.get_latest_tick().mid_price_of_one_sat()
```

then we don't have to duplicate all the use case functions 3 times.

price-server/src/app/mod.rs
```diff
@@ -8,7 +8,7 @@ pub use error::*;
 use shared::{currency::*, payload::OkexBtcUsdSwapPricePayload, pubsub::*};
 
 pub struct PriceApp {
-    price_cache: ExchangePriceCache,
+    pub price_cache: ExchangePriceCache,
```
bodymindarts, line 11: Why is this made public?

price-server/src/exchange_price_cache.rs
```diff
@@ -30,23 +29,22 @@ impl ExchangePriceCache {
...
     timestamp: TimeStamp,
-    ask_price_of_one_sat: UsdCents,
-    bid_price_of_one_sat: UsdCents,
+    pub ask_price_of_one_sat: UsdCents,
+    pub bid_price_of_one_sat: UsdCents,
+}
+
+impl BtcSatTick {
+    pub fn mid_price_of_one_sat(&self) -> UsdCents {
+        (self.bid_price_of_one_sat.clone() + self.ask_price_of_one_sat.clone()) / 2
```
bodymindarts, line 46: Too much `.clone()`ing. You need to `impl Add` on the pointer type `&UsdCents`.
Something like:
```
        impl std::ops::Add<&$name> for &$name {
            type Output = $name;

            fn add(self, rhs: &$name) -> Self::Output {
                Self {
                    inner: self.inner + res
                }
            }
        }

```

shared/src/currency/mod.rs
```diff
@@ -50,6 +50,26 @@ macro_rules! currency {
             }
         }
 
+        impl std::ops::Add<$name> for $name {
```
bodymindarts, line 53: Change this to be implemented on the pointer type `&$name` or add it.

shared/src/payload/primitives.rs
```diff
@@ -38,6 +46,19 @@ impl PriceRatioRaw {
     }
 }
 
+pub struct ConvertU64ToF64 {
```
bodymindarts, line 49: This is a pointless struct. It doesn't add anything you can't do with stdlib - delete it.

price-server/src/server/mod.rs
```diff
@@ -143,7 +143,12 @@ impl Price for PriceService {
             .app
             .get_cents_per_sats_exchange_mid_rate(Sats::from_major(req.amount_in_satoshis))
             .await?;
-        let ratio_in_cents_per_satoshis = f64::from_bits(mid_amount_in_cents);
+        let ratio_in_cents_per_satoshis = ConvertU64ToF64 {
```
bodymindarts, line 146: This struct for converting u64 -> f64 is not appropriate. Use stdlib or a cast.

shared/src/currency/mod.rs
```diff
@@ -59,7 +79,6 @@ macro_rules! currency {
         }
     };
 }
-
```
bodymindarts, line 62: why delete this line break?

price-server/src/server/mod.rs
tvpeter: extracted to a new fn and unit tested

price-server/src/server/mod.rs
```diff
@@ -135,9 +135,23 @@ impl Price for PriceService {
...
         &self,
-        _request: Request<GetCentsPerSatsExchangeMidRateRequest>,
+        request: Request<GetCentsPerSatsExchangeMidRateRequest>,
     ) -> Result<Response<GetCentsPerSatsExchangeMidRateResponse>, Status> {
-        Err(Status::unimplemented(""))
+        let req = request.into_inner();
+
+        let mid_amount_in_cents = self
+            .app
+            .get_cents_per_sats_exchange_mid_rate(Sats::from_major(req.amount_in_satoshis))
```
bodymindarts, line 144: I'm confused about this... the proto definition says it doesn't have a field:
```
message GetCentsPerSatsExchangeMidRateRequest {}
message GetCentsPerSatsExchangeMidRateResponse {
  double ratio_in_cents_per_satoshis = 1;
}
```
How does this compile?

price-server/src/server/mod.rs
bodymindarts: It should simply return the price of 1 sat as double. That implies a conversion from `impl From<UsdCent> for f64`.

bodymindarts (changes requested):

price-server/src/exchange_price_cache.rs
tvpeter: I have a blocker implementing `Add` for the pointer type `&$name`. I will need assistance on this. Thank you.

shared/src/payload/primitives.rs
```diff
@@ -9,6 +11,12 @@ crate::string_wrapper! { CurrencyRaw }
 
 const PRICE_IN_CENTS_PRECISION: u32 = 12;
 
+#[derive(Error, Debug)]
+pub enum ConvertU64ToF64Error {
```
bodymindarts, line 15: think this error can be deleted

CHANGELOG.md
```diff
@@ -76,7 +76,7 @@
 - More price-server scafolding
 - Return pined stream from subscribe
 - Clippy
-- Commit Cargo.lock as we are buiding a binary
+- Commit Cargo.lock as we are building a binary
```
bodymindarts, line 79: remove this correction - I've taken `CHANGELOG.md` out of the spell checking.

bodymindarts (approved):

PR #45, opened 2022-08-09
Price fee calculator

tvpeter:
- add price-fee calculator 
- unit test for immediate and delay fees
related: https://github.com/GaloyMoney/stablesats-rs/pull/35


CHANGELOG.md
```diff
@@ -76,7 +76,7 @@
 - More price-server scafolding
 - Return pined stream from subscribe
 - Clippy
-- Commit Cargo.lock as we are buiding a binary
+- Commit Cargo.lock as we are building a binary
```
bodymindarts, line 79: You can remove this - rebase against main where I have removed `CHANGELOG.md` from the spell checking.

.envrc
```diff
@@ -1 +1,5 @@
 export PATH=${PATH}:./target/debug
+export BASE_FEE_RATE=0.005
```
bodymindarts, line 2: This is not where to document these values.

We only want to use env vars to inject sensitive data. Everything else goes into `stablesats.yml` - though these are defaults that should be hard coded in the code.

price-server/src/app/fees_calculator.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl Default for FeeCalculator {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl FeeCalculator {
+    pub fn new() -> Self {
```
bodymindarts, line 29: Config should be injected -> don't want a dependency on the environment (or any outside state) in our domain layer.

bodymindarts (commented):
Its not bad - there are 2 subtle issues.
- config injection
- domain boundary
lets talk about it in person

bodymindarts (approved):

PR #47, opened 2022-08-10
chore: add sats to cents converter

tvpeter:
- add sats to cents converter
- add cents to Usd converter


shared/src/currency/mod.rs
```diff
@@ -50,6 +50,16 @@ macro_rules! currency {
             }
         }
 
+        impl std::ops::Add<$name> for $name {
```
tvpeter, line 53: I re-added this fn because it failed to compile on my local

PR #67, opened 2022-08-17
feat: fetch deposit state and replace f64 values

enigbe:
## What this PR does?
- fetches the deposit status given an `address` and the `amount` deposited
- fix: replaces floating point (f64) parameters with `Decimal`
- fix: reads btc address and amount from environment variables


notes
```diff
@@ -26,7 +26,7 @@ Exchange Funder (deps: OKEX, GALOY)
 -------------------------
 Trader (deps: OKEX)
 -> HedgingNeed
-=> moves money to/from trading account ( controling collateral)
+=> moves money to/from trading account ( controlling collateral)
```
bodymindarts, line 29: did you accidentally commit your notes? Might want to add this to your global gitignore.

okex-client/tests/client.rs
```diff
@@ -47,7 +49,7 @@ async fn client_is_missing_header() -> anyhow::Result<()> {
 
 #[tokio::test]
 async fn transfer_funding_to_trading() -> anyhow::Result<()> {
-    let amount = 0.00001;
+    let amount = Decimal::new(1, 5);
```
bodymindarts, line 52: lets use from `from_str` or `dec!` micro for creating hard coded decimals in tests. I think that's more readable than the `(1, 5)` format

bodymindarts (commented):
I will rebase my rate limit PR onto this

okex-client/tests/client.rs
enigbe: I had initially used `from_str`, but thought `new` was much cleaner. I'll revert.

bodymindarts (approved):

PR #68, opened 2022-08-18
feat: open, get, and close position(s) on/of an instrument

enigbe:
## What this PR does?
- features: Adds methods to place an order (open position), get position, and close all positions of an instrument
- fixes/refactor: Uses the rust_decimal_macros `dec!` to create `Decimal` values

Note:
- The demo API keys need to set. I have a condition that "skips" the tests if they are not set.


okex-client/tests/client.rs
```diff
@@ -60,7 +75,7 @@ async fn funding_account_balance() -> anyhow::Result<()> {
 #[tokio::test]
 async fn trading_account_balance() -> anyhow::Result<()> {
     let avail_balance = configured_okex_client().trading_account_balance().await?;
-    let minimum_balance = Decimal::new(0, 2);
+    let minimum_balance = dec!(0.00);
```
bodymindarts, line 78: why the `00` digits after the decimal?

okex-client/tests/client.rs
```diff
@@ -136,3 +151,83 @@ async fn transfer_state() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn place_order() -> anyhow::Result<()> {
+    if let (Ok(_), Ok(_), Ok(_)) = (
```
bodymindarts, line 157: This makes no sense - just have `demo_okex_client` return `Option<OkexClient>`

price-server/src/fee_calculator/config.rs
```diff
@@ -36,21 +37,11 @@ impl Default for FeeCalculatorConfig {
...
-        );
-        assert_eq!(
-            default_immediate_fee_rate(),
-            Decimal::from_str("0.0005").unwrap()
-        );
-        assert_eq!(
-            default_delayed_fee_rate(),
-            Decimal::from_str("0.0007").unwrap()
-        );
+        assert_eq!(default_base_fee_rate(), dec!(0.0005));
```
bodymindarts, line 43: This test is no longer any good since it is just restating the implementation. Perhaps use `from_str` here on the right side.

okex-client/src/client/mod.rs
```diff
@@ -294,6 +304,101 @@ impl OkexClient {
...
+    /// Open a hedging position on an instrument
+    ///
+    /// Parameters:
+    ///     inst_id(String): instrument ID e.g. "BTC-USD-SWAP" or "BTC-USD"
+    ///     trade_mode(String): "cross"
+    ///     side(String): "buy" or "sell"
+    ///     pos_side(String): "long" or "short"
+    ///     order_type(String): "market"
+    ///     size(u64): e.g. 20
+    pub async fn place_order(
```
bodymindarts, line 316: Need to clean up these arguments into enums instead of `String`s... can do it now or later when we have a calling code that actually uses it.

bodymindarts (commented):
Not bad overall - just some details.

okex-client/src/client/mod.rs
```diff
@@ -374,6 +480,13 @@ impl OkexClient {
             "OK-ACCESS-PASSPHRASE",
             HeaderValue::from_str(self.config.passphrase.as_str())?,
         );
+
+        // Add 'simulated-trading' header flag if environment variable exist
+        let demo_trading_flag = std::env::var("OKEX_SIMULATED_TRADING");
```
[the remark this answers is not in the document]
enigbe: I have resolved this by passing the variable to a constructor.

okex-client/src/client/mod.rs
```diff
@@ -354,30 +407,44 @@ impl OkexClient {
...
-        pos_side: String,
-        margin_mode: String,
+        inst_id: OkexInstrumentId,
+        pos_side: OkexPositionSide,
+        margin_mode: OkexMarginMode,
         ccy: String,
         auto_cxl: bool,
     ) -> Result<ClosePositionData, OkexClientError> {
+        let instrument = match inst_id {
+            OkexInstrumentId::BtcUsdSwap => "BTC-USD-SWAP".to_string(),
```
bodymindarts, line 433: better to implement `Display` and then call `.to_string()`... goes for all of the enums that have fixed string values.

okex-client/src/client/okex_response.rs
```diff
@@ -70,7 +70,7 @@ pub struct TradingBalanceData {
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct TradingBalanceDetails {
-    pub avail_bal: Decimal,
+    pub avail_bal: String,
```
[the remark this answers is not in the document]
enigbe: Unfortunately, `serde` seems unable to de-serialize an empty string into an `Option<Decimal>`.

okex-client/tests/client.rs
enigbe: Fixed

okex-client/tests/client.rs
```diff
@@ -154,80 +156,71 @@ async fn transfer_state() -> anyhow::Result<()> {
...
-            .place_order(
-                "BTC-USD-SWAP".to_string(),
-                "cross".to_string(),
-                "buy".to_string(),
-                "long".to_string(),
-                "market".to_string(),
-                1,
-            )
-            .await?;
+    std::env::set_var("OKEX_SIMULATED_TRADING", "1");
```
bodymindarts, line 159: still being set as env var

okex-client/tests/client.rs
```diff
@@ -154,80 +156,71 @@ async fn transfer_state() -> anyhow::Result<()> {
...
+
+    let client = demo_okex_client();
+    let instrument = OkexInstrumentId::BtcUsdSwap;
+    let margin = OkexMarginMode::Cross;
+    let position_side = OkexPositionSide::long();
```
bodymindarts, line 164: don't use the method, use the `::Long` syntax... get around the string issue by implementing `Display`

okex-client/tests/client.rs
```diff
@@ -154,80 +156,71 @@ async fn transfer_state() -> anyhow::Result<()> {
...
+    let order_side = OkexOrderSide::Buy;
+
+    let order_id = client
+        .place_order(
+            instrument,
+            margin,
+            order_side,
+            position_side,
+            "market".to_string(),
```
bodymindarts, line 173: this one is probably also an enum or?

okex-client/tests/client.rs
```diff
@@ -4,32 +4,33 @@ use rust_decimal::Decimal;
...
 }
 
 fn demo_okex_client() -> OkexClient {
-    // Set environment variable for "x-simulated-trading"
-    env::set_var("OKEX_SIMULATED_TRADING", "1");
-
     let api_key = env::var("OKEX_DEMO_API_KEY").expect("OKEX_DEMO_API_KEY not set");
     let passphrase = env::var("OKEX_DEMO_PASSPHRASE").expect("OKEX_DEMO_PASSPHRASE not set");
     let secret_key = env::var("OKEX_DEMO_SECRET_KEY").expect("OKEX_DEMO_SECRET_KEY not set");
+    let simulated = env::var("OKEX_SIMULATED_TRADING").expect("OKEX_SIMULATED_TRADING not set");
```
bodymindarts, line 28: Do we ever want to run the tests unsimulated? Why are we reading this setting from env and not hard coding it in the test?

okex-client/src/client/mod.rs
```diff
@@ -67,10 +67,49 @@ pub struct PositionId {
...
+pub enum OkexOrderSide {
+    Buy,
+    Sell,
+}
+
 pub struct OkexClientConfig {
     pub api_key: String,
     pub passphrase: String,
     pub secret_key: String,
+    pub simulated: String,
```
bodymindarts, line 112: shouldn't this be a boolean? How do callers know what values are accepted here?

okex-client/tests/client.rs
enigbe: I think that it is better to have `demo_okex_client` return `OkexClient` and to set the environment variables on github. This way,  the affected tests run and pass only when the variables have been set. I got rid of the `if` block.

okex-client/src/client/mod.rs
enigbe: I have created four enums: `OkexInstrumentId`, `OkexMarginMode`, `OkexOrderSide`, and `OkexPositionSide` to address this.

okex-client/tests/client.rs
enigbe: For majority of the tests, we can run them simulated against the demo API but not against endpoints to deposit or withdraw. Having removed the withdraw test, there is no need to run tests unsimulated. 

With regards to reading variables from the environment, I could easily hardcode them in the test but then the tests will be linked to my account's API keys. If I, for whatever reason, delete those keys, all tests will fail. If the demo API keys for galoy's OKX account are shared with me, then reading from the environment would not be needed.

okex-client/src/client/mod.rs
enigbe: Fixed

okex-client/tests/client.rs
enigbe: This has been removed.

okex-client/src/client/mod.rs
```diff
@@ -57,10 +58,110 @@ pub struct DepositStatus {
...
+}
+
+#[derive(Debug, Clone)]
+pub enum OkexPosition {
+    Long,
+    Short,
+}
+
+#[derive(Debug, Clone)]
+pub enum OkexPositionSide {
```
bodymindarts, line 108: Position Side `Long/Short` => always in `net` mode (check in constructor?)

okex-client/src/client/mod.rs
```diff
@@ -355,6 +545,7 @@ impl OkexClient {
         pre_hash: String,
     ) -> Result<HeaderMap, OkexClientError> {
         let sign_base64 = self.sign_okex_request(pre_hash);
+        let simulation_flag = self.config.simulated as i32;
```
bodymindarts, line 548: too implicit `i32::from(bool)`?

okex-client/tests/client.rs
```diff
@@ -1,20 +1,37 @@
 use std::env;
 
 use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
 
 use okex_client::{
-    OkexClient, OkexClientConfig, OkexClientError, OKEX_MINIMUM_WITHDRAWAL_AMOUNT,
+    OkexClient, OkexClientConfig, OkexClientError, OkexInstrumentId, OkexMarginMode, OkexOrderSide,
```
bodymindarts, line 7: `use okex_client::*`

okex-client/tests/client.rs
```diff
@@ -1,20 +1,37 @@
...
         secret_key,
+        simulated,
+    })
+}
+
+fn demo_okex_client() -> OkexClient {
+    let api_key = "1f360c35-4942-489b-a22c-3e38741de501".to_string();
+    let passphrase = "@Stablesats123".to_string();
+    let secret_key = "BD6CA6849F3A2CACA16F62E3D354D61C".to_string();
+    let simulated = true;
```
bodymindarts, line 29: only hard code simulated

okex-price/src/price_feed/tick.rs
```diff
@@ -5,6 +5,7 @@ use shared::time::*;
 
 #[derive(Clone, Deserialize, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
+#[allow(clippy::derive_partial_eq_without_eq)]
```
bodymindarts, line 8: lets just defer to clippy as default

bodymindarts (approved):

PR #71, opened 2022-08-20
feat: configure position_mode and position_side in constructor

enigbe:
## What this PR does?
- adds `position_mode` (`long_short_mode` vs `net_mode`) and `position_side` (`long`/`short`/`net`) to the `OkexClient` constructor. The latter is used as a type guard to ensure that client calls to the `place_order` and `close_positions` API pass in the acceptable side (`net`). 
- Clean-up (clippy): Derives Eq where PartialEq is derived
- Replace type casting of simulated configuration parameter with explicit type conversion

Note:
I started working on a method to get the current account configuration to ensure that the `position_mode` is the same as in the configuration, i.e. `net_mode`. This method could be used in the call to place an order but would mean an extra API call will be made to OKX, which is undesirable given the rate limits on their API.

### DIFF
--- a/okex-client/src/client/error.rs
+++ b/okex-client/src/client/error.rs
@@ -12,4 +12,6 @@ pub enum OkexClientError {
     UnexpectedResponse { msg: String, code: String },
     #[error("OkexClientError: {0}")]
     DecimalConversion(#[from] rust_decimal::Error),
+    #[error("OkexClientError: {code:?} - {msg:?}")]
+    PositionSide { msg: String, code: String },
 }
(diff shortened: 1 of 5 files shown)


bodymindarts:
> adds position_mode (long_short_mode vs net_mode) and position_side (long/short/net) to the OkexClient constructor. The latter is used as a type guard to ensure that client calls to the place_order and close_positions API pass in the acceptable side (net).

This is the opposite of what we discussed. We do _not_ want to expose this config, since our use cases all assume we are using `net`. We should verify the assumption that the account is setup correctly (ie to `net`) in the constructor (and error otherwise) but then not expose any of these options externally. They should be (hard coded) internal details.

bodymindarts:
Merging this because it fixes the test failure in `open_position` but please address ^^ in another PR.

closes https://github.com/GaloyMoney/stablesats-rs/issues/73

PR #78, opened 2022-08-22
fix: construct okex client if position mode is set to "net_mode"

enigbe:
## What this PR does
- constructs  `OkexClient` only if the position mode (i.e. order placement) is set to `"net_mode"`
- adds position_side of `"net"`, order_type of `"market"`, and margin_mode of `"cross"` to config

## Context
- Related PR: #71 
- Addresses this [comment](https://github.com/GaloyMoney/stablesats-rs/pull/71#issuecomment-1221382482)

### DIFF
--- a/okex-client/src/client/mod.rs
+++ b/okex-client/src/client/mod.rs
@@ -34,11 +34,38 @@ pub struct OkexClient {
 }
 
 impl OkexClient {
-    pub fn new(config: OkexClientConfig) -> Self {
+    fn new(config: OkexClientConfig) -> Self {
         Self {
             client: ReqwestClient::new(),
             config,
         }
+    }
+
+    pub async fn create(config: OkexClientConfig) -> Result<Self, OkexClientError> {
+        // 1. Get account configuration
+        let client = Self::new(config);
+        let path = "/api/v5/account/config";
+        let config_url = Self::url_for_path(path);
+        let headers = client.get_request_headers(path)?;
+
+        let response = client
+            .rate_limit_client(path)
+            .await
+            .get(config_url)
+            .headers(headers)
+            .send()
+            .await?;
+        let config_data =
+            Self::extract_response_data::<OkexAccountConfigurationData>(response).await?;
+
+        // 2. Check postion_mode, i.e. order placement mode
+        if config_data.pos_mode == *"net_mode" {
+            return Ok(client);
+        }
+        Err(OkexClientError::PositionMode {
+            msg: format!("Expected `net_mode`, got {}", config_data.pos_mode),
+            code: "0".to_string(),
+        })
     }
 
     pub async fn rate_limit_client(&self, key: &'static str) -> &ReqwestClient {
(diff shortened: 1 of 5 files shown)


okex-client/src/client/mod.rs
```diff
@@ -34,13 +34,40 @@ pub struct OkexClient {
...
+    fn new(config: OkexClientConfig) -> Self {
         Self {
             client: ReqwestClient::new(),
             config,
         }
     }
 
+    pub async fn create(config: OkexClientConfig) -> Result<Self, OkexClientError> {
+        // 1. Get account configuration
+        let client = Self::new(config);
```
bodymindarts, line 46: There is no need for `new` and `create`. Just rename `create` to `new` and inline what is now `new` here

okex-client/src/client/primitives.rs
```diff
@@ -147,4 +147,6 @@ pub struct OkexClientConfig {
     pub simulated: bool,
     pub position_mode: OkexPositionMode,
     pub position_side: OkexPositionSide,
+    pub order_type: OkexOrderType,
```
bodymindarts, line 150: I find it very confusing to add these settings here.

- are they static? ie will they be the same on every call to `place_order` or `close_position`? If so we should not expose them as config options but hard code it somewhere.
- might they change from one call to `place_order` to the next? If so then they should be passed as function arguments.

okex-client/tests/client.rs
```diff
@@ -205,38 +225,22 @@ async fn get_positions() -> anyhow::Result<()> {
...
 #[tokio::test]
 async fn close_positions() -> anyhow::Result<()> {
-    if let Some(client) = demo_okex_client() {
+    if let Some(client) = demo_okex_client().await? {
         let instrument = OkexInstrumentId::BtcUsdSwap;
-        let margin = OkexMarginMode::Cross;
-        let position_side = OkexPositionSide::Net;
         let order_side = OkexOrderSide::Buy;
-        let order_type = OkexOrderType::Market;
+        let currency = "BTC".to_string();
```
bodymindarts, line 231: Will this every be a different currency?

- if we are always trading with `BTC` then we should not expose it in the interface.

okex-client/src/client/mod.rs
```diff
@@ -34,13 +34,40 @@ pub struct OkexClient {
...
+            .send()
+            .await?;
+        let config_data =
+            Self::extract_response_data::<OkexAccountConfigurationData>(response).await?;
+
+        // 2. Check position_mode, i.e. order placement mode
+        if config_data.pos_mode == *"net_mode" {
+            return Ok(client);
+        }
+        Err(OkexClientError::PositionMode {
```
bodymindarts, line 65: Can we call this `MissconfiguredAccount` or something like that instead of `PositionMode`?

Also shouldn't have `msg, code` format as that is used for the responses we pass on that come directly from the API.
Perhaps just `MissconfiguredAccount(string)` or something like that.

bodymindarts:
Merging because tests are passing... follow up with another PR (to not diverge from main too much).

okex-client/src/client/primitives.rs
enigbe: I had a chat with @sebastienverreault about these and he provided some clarification. 

1. For the `position_side`, it has to be `net`, given that the `position_mode` is `net`. The OKX documentation highlights it [here](https://www.okx.com/docs-v5/en/#rest-api-trade-place-order) (check `posSide` ).
2. For the `order_type`, it's `market` that we use. 

I don't think that these will change except we change our strategy. Given that some of these config parameters are shared in the call to `place_order` and `close_position`, it seemed better to create them once (in the config).

bodymindarts: We probably have 2 types of config. One that we input at runtime and one that remains static / hard coded.

PR #79, opened 2022-08-22
feat: remove margin_mode, position_side, order_type from config

enigbe:
## What does this PR do?
- Remove `margin_mode`, `position_side`, and `order_type` from OkexClient config, alongside code cleanup
- Related: #78


bodymindarts (approved):
Awesome - I think this is probably as far as we can go with this crate without knowing how / where we're going to use it.

Unless there is an important endpoint still unimplemented I would move on to something else.

PR #81, opened 2022-08-23
refactor: extract balance update into single query

bodymindarts:
Simplified some application code into a single query:
```
 WITH buy_amounts AS (
    SELECT MAX(idx) as max_buy, buy_unit, SUM(buy_amount) AS to_sub FROM user_trades WHERE idx > $1 GROUP BY buy_unit
  ),
  sell_amounts AS (
    SELECT MAX(idx) as max_sell, sell_unit, SUM(sell_amount) AS to_add FROM user_trades WHERE idx > $1 GROUP BY sell_unit
  )
 SELECT
   unit AS "unit: _",
   current_balance + COALESCE(to_add, 0) - COALESCE(to_sub, 0) AS new_balance,
   MAX(GREATEST(max_buy, max_sell)) OVER () AS new_latest_idx
 FROM user_trade_balances
 LEFT JOIN buy_amounts ON unit=buy_unit
 LEFT JOIN sell_amounts ON unit=sell_unit;
 ```
 
 but is this readable for you @enigbe @sebastienverreault ?
 
 It is about 50% less code and more performant to do it like this but you have to know your SQL.
 
 WDYT is using SQL directly a style we want to go with? I would be in favour but want your input.

### DIFF
--- a/user-trades/sqlx-data.json
+++ b/user-trades/sqlx-data.json
@@ -44,130 +44,6 @@
     },
     "query": "INSERT INTO user_trades (uuid, buy_unit, buy_amount, sell_unit, sell_amount) VALUES ($1, $2, $3, $4, $5) RETURNING idx"
   },
-  "27702793a9046cc0abd9157698cb260292eabe86d667d703af5cf52439a21ca9": {
-    "describe": {
-      "columns": [
-        {
-          "name": "unit: _",
-          "ordinal": 0,
-          "type_info": {
-            "Custom": {
-              "kind": {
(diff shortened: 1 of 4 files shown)


enigbe (approved):

enigbe (approved):
The SQL code is fairly readable. I can always review a textbook if some part of the query is unclear. Also, using SQL directly is a style I have seen used in [Zero To Production In Rust](https://algoluca.gumroad.com/l/zero2prod) so I personally do not mind it.

PR #83, opened 2022-08-25
feat: galoy wallet library

enigbe:
## What this PR does?
- Adds `GaloyClient` to query the galoy graphql API.
- Adds methods to get an authentication code (linked to phone number), to login to wallet account, to retrieve BTC and USD transactions, and to get the default wallets balances


galoy-wallet/src/client/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Debug, Clone)]
+pub struct GaloyClient {
+    client: ReqwestClient,
+    config: GaloyClientConfig,
+}
+
+#[derive(Debug, Clone)]
+pub struct GaloyClientConfig {
+    pub api: String,
+    pub phone_number: String,
```
nicolasburtey, line 24: we probably won't store the phone number but more likely the JWT, specially for the dealer.

enigbe: I will eventually add the JWT, but I have no need for it yet until I can get the `authentication_code()` and `login()` functions to work with my local environment for galoy backend.

bodymindarts: The current dealer gets phone/code injected and creates its own auth token - so we are using the same model here. It means less manual steps for provisioning Stablesats in a live environment.

galoy-client/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "galoy-client"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+futures = "0.3.23"
+graphql_client = {version = "0.11.0", features = ["reqwest-blocking", "reqwest"]}
```
bodymindarts, line 10: Why blocking?

galoy-client/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+reqwest = {version = "0.11.11", features = ["json"]}
+serde = "1.0.144"
+thiserror = "1.0.32"
+tokio = {version = "1.20.1", features = ["full"]}
```
bodymindarts, line 14: why full?

galoy-client/src/client/error.rs
```diff
@@ -0,0 +1,15 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum GaloyClientError {
```
bodymindarts, line 4: Lets start using the pattern that every crate has an error in root (instead of in sub-modules).

So move this error under `src/error`. The only exception would be for shared where there can be errors in submodules. Need to do a small refactoring to make this consistent in all sub-crates.

galoy-client/src/client/graphql/queries/transactions_list.graphql
```diff
@@ -0,0 +1,51 @@
+query StablesatsTransactionsList(
+  $first: Int
+  $last: Int
+  $before: String
+  $after: String
+) {
+  me {
+    defaultAccount {
+      __typename
+      transactions(first: $first, last: $last, before: $before, after: $after) {
```
bodymindarts, line 10: do we need all of these arguments?

galoy-client/src/client/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+            .build()?;
+
+        Ok(Self {
+            client,
+            config,
+            jwt,
+        })
+    }
+
+    pub async fn authentication_code(
```
bodymindarts, line 79: When do we need this? To login we aways already have the code right?

galoy-client/src/client/graphql/queries/transactions_list.graphql
bodymindarts: In the code I can only see the `after` being used.

galoy-client/src/client/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+                ))
+            }
+        };
+
+        let ccy_tx: Vec<stablesats_transactions_list::StablesatsTransactionsListMeDefaultAccountTransactionsEdges> = tx
+            .into_iter()
+            .filter(move |transaction| transaction.node.settlement_currency == wallet_currency)
+            .collect();
+
+        Ok(Box::pin(stream::iter(ccy_tx)))
```
bodymindarts, line 192: I see that you return a stream like I suggested but I see now I made a mistake in my thinking... just return the Vec.

galoy-client/src/client/mod.rs
```diff
@@ -0,0 +1,361 @@
...
+        if response_data.is_none() {
+            return Err(GaloyClientError::UnknownResponse(
+                "Empty `data` in response".to_string(),
+            ));
+        }
+
+        let auth_token = match response_data {
+            Some(login_data) => StablesatsLogin::try_from(login_data)?.auth_token,
+            None => {
+                return Err(GaloyClientError::UnknownResponse(format!(
```
bodymindarts, line 124: What is the difference between `UnknownResponse` and `GraphQLApi`? You seem to be using both of those errors in the same way.

galoy-client/src/client/queries.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "src/client/graphql/schema.graphql",
+    query_path = "src/client/graphql/queries/transactions_list.graphql",
+    response_derives = "Debug, PartialEq, Clone"
+)]
+pub struct StablesatsTransactionsList;
+pub type WalletId = String;
+pub type Timestamp = u64;
```
bodymindarts, line 66: Can this be deserialised to a chrono type?

galoy-client/src/client/queries.rs
```diff
@@ -0,0 +1,209 @@
...
+pub type Memo = String;
+pub(crate) type SignedAmount = f64;
```
bodymindarts, line 68: please use `Decimal` for all money amounts. Float should NEVER be used for money! It is imprecise.

galoy-client/src/client/queries.rs
```diff
@@ -0,0 +1,209 @@
...
+        let wallets = default_account.wallets;
+
+        let btc_wallet = wallets
+            .clone()
+            .into_iter()
+            .find(|wallet| wallet.wallet_currency == WalletCurrency::BTC);
+
+        let usd_wallet = wallets
+            .into_iter()
+            .find(|wallet| wallet.wallet_currency == WalletCurrency::USD);
```
bodymindarts, line 140: I think for this you could use [partition_map](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_map) from the itertools crate. Then you don't need to `clone`.

galoy-client/src/client/queries.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[derive(GraphQLQuery)]
+#[graphql(
+    schema_path = "src/client/graphql/schema.graphql",
+    query_path = "src/client/graphql/queries/onchain_tx_fee.graphql",
+    response_derives = "Debug"
+)]
+pub struct StablesatsOnChainTxFee;
+pub type TargetConfirmations = u32;
+pub type SatAmount = u128;
```
bodymindarts, line 157: `Decimal`

galoy-client/src/client/mod.rs
nicolasburtey: I think I should probably switch to using this library for galoy-cli?

in that case this would be useful for now

bodymindarts: I've thought about this and think they should remain separate.

- different use cases
- different lifecycles
- different teams

I don't see any upside for unifying atm. But we can discuss at a standup if you have a strong opinion but for me this is a classic case of DRY considered harmful. What is the trigger for unifying when its just going to create friction?

nicolasburtey: I need in the cli things like:
- get/create address
- get list of transactions
- get balance
which has all been done here already

I could for sure copy the code or redo it, but the use case seems close enough to me that using a single library seems to make sense.

Things like batching that are specific to cli would not be in the library but on top of it.

bodymindarts: Its the `close enough` that I am worried about here. If I were in your position I would copy / paste.
But if you feel you'll be better served by depending on this sub-crate go ahead. Just be aware that there are likely to be breaking changes as we start using it in client code and know more about the needs.

In particular I could easily see the return types and arguments and initialisation to go through some iteration before were at a point that it stabilises for the use cases in this repo.

Not being able to iterate liberally (ie. needing to take external use cases into account and avoid breaking changes) causes coordination overhead that will slow everyone down.

nicolasburtey: ok fair. I don't have to rush to use this library. I can do it in a couple of weeks once stablesats v2 is in prod.

galoy-client/src/client/queries.rs
enigbe: I have changed this but I think that `SatAmount` should be an unsigned integer. With `Decimal`, we run the risk of validating fractional sat values.

galoy-client/src/client/mod.rs
enigbe: Removed the redundant `UnknownResponse`

galoy-client/src/client/mod.rs
enigbe: This now returns a vector of transactions

galoy-client/Cargo.toml
enigbe: Without the feature flag of `full` or `macros`, tokio test was not found. I have set the feature to `macros` so that it compiles. Uncertain why this is the case given that some other crates do not have any features flags.

galoy-client/Cargo.toml
enigbe: Please disregard. I had initially implemented a blocking `GaloyClient::connect()` method but refactored it to be async because blocking calls panic in an async runtime. Feature has been removed.

galoy-client/src/client/graphql/queries/transactions_list.graphql
bodymindarts: hmm okay don't want any warnings

galoy-client/src/client/queries.rs
bodymindarts: Thats a good point, do that for now - will think about it some more and we'll see what we really need when we start using it.

bodymindarts (approved):

PR #91, opened 2022-09-02
feat hedging

bodymindarts:


bodymindarts:
The hedging test is working now e2e. Still many details to fix before production release but I think its worth merging and iterating on it in main.

.cargo/audit.toml
```diff
@@ -0,0 +1,2 @@
+[advisories]
+ignore = ["RUSTSEC-2020-0071"]
```
nicolasburtey, line 2: why is that necessary?

bodymindarts: there is a feature in `chrono` called `time` that is deprecated. The latest version of `chrono` removes it completely but its not fully released yet.
When its active it has an issue that a crate that gets brought in can segfault under certain circumstances.
We don't need that feature but unfortunately its being brought in transitively by `sqlxmq` so we cannot turn it off.
As we are not using any code paths from that feature it shouldn't be an issue and once a chrono crate gets its newest release it shouldn't be an issue anyway.

PR #98, opened 2022-09-06
Feat hedging hysteresis

sebastienverreault:
Change hedging algo to behave like Dealer v1


okex-client/src/client/mod.rs
```diff
@@ -347,21 +347,14 @@ impl OkexClient {
             .send()
             .await?;
 
-        let PositionData {
-            notional_usd, pos, ..
-        } = Self::extract_response_data::<PositionData>(response).await?;
+        let PositionData { notional_usd, .. } =
+            Self::extract_response_data::<PositionData>(response).await?;
 
-        let direction = pos.parse::<Decimal>().unwrap_or(Decimal::ZERO);
```
bodymindarts, line 354: This is dangerous - if we don't distinguish between exposure direction the dealer might be in a completely opposite state and the system would not know it!

bodymindarts (changes requested):
Looks good in general but 2 comments:
- I oppose reporting `abs_exposure` as this has some serious risk!
- you have significantly changed the logic in `adjustment_action` but not changed / added any tests. There is no indication that this change was even needed to satisfy the previous tests. I count at least 7 different branches - so it is impossible to test all paths with 5 tests. Please update the tests to better reflect (ie. document) the logic. What do the constants actually signify etc? Perhaps you can express the tests with small deltas from the constants or something like that.

okex-client/src/client/mod.rs
sebastienverreault: Right and if possible to happen, I thought we should validate it at the source instead of all the clutter it generates downstream.

bodymindarts: It is possible to happen and the business logic has to be aware of it. How else can it make the right decisions if it doesn't know what state the position is in?

bodymindarts: > Right and if possible to happen, I thought we should validate it at the source instead of all the clutter it generates downstream.

Let me clarify something here `validation` is something we do to input that we can reject. EG if a user provides us a value that is out of the range of the bounds of what we accept we can and should validate it and reject if it doesn't fulfil our implicit contract.
This value is not something we can validate as it is not something we can reject - it represents external state that simply _is what it is_ and we have to deal with it no matter what.

sebastienverreault:
Now adding new integration test

PR #107, opened 2022-09-13
Test hedging hysteresis

sebastienverreault:
add hedging hysteresis unit tests
fix minimum liability threshold 
add minimum liability threshold unit tests
add dealer v1 scenario based integration test


hedging/tests/fixtures/hedging.json
```diff
@@ -10,4 +10,4 @@
       "liability": "0"
     }
   ]
-}
+}
```
bodymindarts, line 13: You need to fix your editor to add new lines at end of file.

hedging/tests/fixtures/scenario_01.json
```diff
@@ -0,0 +1,40 @@
...
+            "liability": "55"
+        },
+        {
+            "liability": "30"
+        },
+        {
+            "liability": "15"
+        }
+    ]
+}
```
bodymindarts, line 40: Missing newline.

hedging/tests/hedging.rs
```diff
@@ -138,3 +139,69 @@ async fn hedging() -> anyhow::Result<()> {
...
+                okex_poll_frequency: std::time::Duration::from_secs(1),
+            },
+            okex_client_config(),
+            pubsub_config,
+        )
+        .await
+    });
+    tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+
+    let expected_exposure = [
```
bodymindarts, line 177: I don't like that this is defined here while the fixtures are in the json blob.
Can we either derive the expectation from the payload or encode them together?
Perhaps:
scenario_01.json
```
{
  scenario: [
  { payload: { liability: 100}
    expectation: { lowerBound: 99, upperBound: 101 }
    },
    <etc>
  ]
}
```

hedging/src/adjustment_action.rs
```diff
@@ -164,10 +164,97 @@ mod tests {
...
+        let liability = Decimal::from(nominal_liability);
+        let exposure = Decimal::from(-nominal_liability) * LOW_BOUND_RATIO_SHORTING;
+
+        let adjustment = determine_action(liability, exposure);
+        assert_eq!(adjustment, AdjustmentAction::DoNothing);
+    }
+
+    #[test]
+    fn low_bound_below() {
+        // condition: exposure / liability < LOW_BOUND_RATIO_SHORTING
```
bodymindarts, line 187: It seems like these tests are simply mirroring the implementation... hmm.

Also I'm not a fan of the comments. I see their value but perhaps they should at least be moved above the fn as doc comments:
https://doc.rust-lang.org/rust-by-example/meta/doc.html
```
/// ...
#[test]
fn ..
```

hedging/tests/hedging.rs
```diff
@@ -114,7 +112,10 @@ async fn hedging() -> anyhow::Result<()> {
     });
     tokio::time::sleep(std::time::Duration::from_secs(3)).await;
 
-    let mut payloads = load_fixture()?.payloads.into_iter();
+    let mut payloads = load_fixture("./tests/fixtures/hedging.json")
```
bodymindarts, line 115: I think we can delete the `hedging` test - `scenario_01` seems to capture everything - so its redundant.

hedging/src/adjustment_action.rs
sebastienverreault: I can do the calculations offline and put magic numbers here, it will be the same "mirroring" of the implementation, no choice else they won't pass. So those unit test will protect from implementation changes breaking them.

Comments were for me before writing the code, I can remove them, we don't have any, anywhere.

What do you think?

hedging/tests/hedging.rs
sebastienverreault: Ok, removing.

hedging/tests/fixtures/hedging.json
sebastienverreault: I see, I was wondering about this... thx

hedging/tests/hedging.rs
sebastienverreault: Yeah wanted to do exactly that, and was going to ask what you thought was the best way...

bodymindarts: I think the best way would be to derive the expectation from the `payload.liability`. If the upper / lower bound is always +/- 1 this should be simple.

hedging/src/adjustment_action.rs
bodymindarts: If you don't need them anymore my preference is to delete the comments.

no magic numbers - I guess the mirroring is okay for now - can't think of anything better.

hedging/tests/fixtures/scenario_01.json
```diff
@@ -0,0 +1,40 @@
+{
+    "payloads": [
```
bodymindarts, line 2: I don't really understand this test. What edge cases / code paths is it going through that the `hedging` test didn't touch?
All things being equal we should prefer the test with the shortest runtime. I'm wondering if this is actually redundant and we keep the original one - or perhaps extend it slightly if there is a code path we missed.

Note that integration tests are not intended to test the exact logic - that's what unit tests are for.

sebastienverreault: All right, in that case I will restore your test and remove this one and we should be good to go.

hedging/src/adjustment_action.rs
sebastienverreault: perfect will do that. thx

bodymindarts:
merged locally

PR #122, opened 2022-09-20
feat(wip): resilient subscription

enigbe:
## What this PR does?
- Adds resilience to redis connection


shared/src/pubsub/subscriber.rs
```diff
@@ -43,4 +44,48 @@ impl Subscriber {
             },
         )))
     }
+
+    pub async fn subscribe_alt<M: MessagePayload>(
+        &self,
+    ) -> Result<Receiver<std::pin::Pin<Box<dyn Stream<Item = Envelope<M>> + Send>>>, SubscriberError>
```
bodymindarts, line 50: the return type should be the same as before. The channel shouldn't transmit the stream but the items.

PR #129, opened 2022-09-26
feat: throttle price tick publishing

enigbe:
Throttles the publishing of price ticks

### DIFF
--- a/shared/src/pubsub/publisher.rs
+++ b/shared/src/pubsub/publisher.rs
@@ -1,9 +1,19 @@
 use fred::prelude::*;
+use std::time::Duration;
 use tracing::instrument;
 
 use super::config::*;
 use super::error::PublisherError;
 use super::message::*;
+
+use governor::{
+    clock::DefaultClock, state::keyed::DefaultKeyedStateStore, Jitter, Quota, RateLimiter,
+};
+use std::num::NonZeroU32;
+
+lazy_static::lazy_static! {
+    static ref LIMITER: RateLimiter<&'static str, DefaultKeyedStateStore<&'static str>, DefaultClock>  = RateLimiter::keyed(Quota::per_minute(NonZeroU32::new(30).unwrap()));
+}
 
 #[derive(Clone)]
 pub struct Publisher {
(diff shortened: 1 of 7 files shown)


shared/src/pubsub/publisher.rs
```diff
@@ -18,12 +30,43 @@ impl Publisher {
...
-        Ok(Self { client })
+        Ok(Self {
+            client,
+            rate_limiter: Arc::new(RateLimiter::keyed(Quota::per_minute(
+                NonZeroU32::new(RATE_LIMIT).unwrap(),
+            ))),
+        })
+    }
+
+    pub fn rate_limit_publisher(
```
bodymindarts, line 41: why pub?

shared/src/pubsub/publisher.rs
```diff
@@ -18,12 +30,43 @@ impl Publisher {
...
+    ) -> Result<bool, PublisherError> {
+        match self.rate_limit_publisher(<P as MessagePayload>::message_type()) {
+            Ok(publisher) => {
+                let publish_result = publisher.publish(payload).await;
+                match publish_result {
+                    Ok(_res) => Ok(true),
+                    Err(err) => Err(err),
+                }
+            }
+            Err(_) => {
```
bodymindarts, line 62:
```suggestion
        if let Ok(_) = self.rate_limiter.check_key(&key) {
        self.publish.await?;
        Ok(true)
        } else {
          Ok(false)
        }
```

shared/src/pubsub/publisher.rs
```diff
@@ -18,12 +30,43 @@ impl Publisher {
...
+        })
+    }
+
+    pub fn rate_limit_publisher(
+        &self,
+        key: &'static str,
+    ) -> Result<&Publisher, NotUntil<QuantaInstant>> {
+        self.rate_limiter.check_key(&key)?;
+        Ok(self)
+    }
```
bodymindarts, line 47:
```suggestion

```

enigbe: Your earlier suggestion was right. This should not be public.

shared/src/pubsub/publisher.rs
```diff
@@ -1,29 +1,68 @@
...
         client
             .wait_for_connect()
             .await
             .map_err(PublisherError::InitialConnection)?;
-        Ok(Self { client })
+
+        Ok(Self {
+            client,
+            rate_limiter: Arc::new(RateLimiter::keyed(
+                Quota::with_period(Duration::from_secs(config.rate_limit_interval.unwrap()))
```
bodymindarts, line 37: how is `unwrap` guaranteed to work?

shared/src/pubsub/publisher.rs
```diff
@@ -1,29 +1,68 @@
...
+                    .unwrap()
+                    .allow_burst(NonZeroU32::new(MAX_BURST).unwrap()),
+            )),
+        })
+    }
+
+    fn rate_limit_publisher(
```
bodymindarts, line 44: What is this function for? Why not inline `self.rate_limiter.check_key(&key)`?

shared/src/pubsub/publisher.rs
```diff
@@ -1,29 +1,68 @@
...
             .wait_for_connect()
             .await
             .map_err(PublisherError::InitialConnection)?;
-        Ok(Self { client })
+
+        Ok(Self {
+            client,
+            rate_limiter: Arc::new(RateLimiter::keyed(
+                Quota::with_period(Duration::from_secs(config.rate_limit_interval.unwrap()))
+                    .unwrap()
```
bodymindarts, line 38: We should never use unwrap in production code - always `expect`

PR #142, opened 2022-10-03
fix: galoy client error return

### LINKED ISSUE #115: galoy client not returning errors
When gql returns an error its not being passed back to the caller.

enigbe:
## What this PR does?
- Introduces `GrapgQLNested` and `GrapgQLTopLevel`  error variants from the Graphql API
- Propagates these errors to `GaloyClientError`

## Related Issue
- fixes #115

### DIFF
--- a/galoy-client/src/error.rs
+++ b/galoy-client/src/error.rs
@@ -1,3 +1,9 @@
+use std::collections::HashMap;
+
+use graphql_client::{Location, PathFragment};
+use serde::Deserialize;
+use serde_json::Value;
+use std::collections::hash_map::RandomState;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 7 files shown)


galoy-client/src/error.rs
```diff
@@ -11,3 +17,30 @@ pub enum GaloyClientError {
     #[error("GaloyClientError - Authentication: {0}")]
     Authentication(String),
 }
+
+#[derive(Debug, Deserialize)]
+pub struct InnerError {
```
bodymindarts, line 22: Why are these individual structs and not part of `GaloyClientError` enum?
I'm thinking the `GraphQLApi(String),` variant should simply be extended to have more fields or something like that.
Or perhaps another variant `GraphQLResponseError` but not individual structs - seems very strange.

PR #145, opened 2022-10-04
fix: inject tracing headers

### LINKED ISSUE #143: issues with galoy client
Headers need to be injected for every call (not just tx list). And need to use rustle.

There are now 2 different request clients being used in `Galoy-client`.
The `inner` one https://github.com/GaloyMoney/stablesats-rs/blob/main/galoy-client/src/client/mod.rs#L35
which is configured to use rustls https://github.com/GaloyMoney/stablesats-rs/blob/main/galoy-client/src/client/mod.rs#L62.
And another one that only gets created during `transactions_list` https://github.com/GaloyMoney/stablesats-rs/blob/b4b882d39d8ebb3f5d62f37816517bbff9fd20a3/galoy-client/src/client/mod.rs#L178-L183.

Obviously we should only have 1 client that we use internally and we need to add the headers to every call. Having 2 different code paths that configures the clients is no good because of the duplication - it has led to stablesats breaking on staging https://ui.honeycomb.io/galoy/datasets/galoy-staging/result/zc1TxpVM2EE

enigbe:
## What this PR does?
- Injects tracing headers in all requests
- Ensures there is only one request client

## Related issue
- Fixes #143

### DIFF
--- a/galoy-client/src/client/mod.rs
+++ b/galoy-client/src/client/mod.rs
@@ -5,10 +5,10 @@ mod queries;
 mod transaction;
 
 use galoy_tracing::*;
-use graphql_client::reqwest::post_graphql;
+use graphql_client::{GraphQLQuery, Response};
 use reqwest::{
-    header::{self, HeaderValue, AUTHORIZATION},
-    Client as ReqwestClient,
+    header::{HeaderValue, AUTHORIZATION},
+    Client as ReqwestClient, Method,
 };
 use rust_decimal::Decimal;
 use tracing::instrument;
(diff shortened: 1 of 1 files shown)


galoy-client/src/client/mod.rs
```diff
@@ -328,27 +308,44 @@ impl GaloyClient {
...
+        let result = response_data.ok_or_else(|| {
+            GaloyClientError::GraphQLApi("Empty `onChainTxFee` in response data".to_string())
+        })?;
 
         let onchain_tx_fee = StablesatsTxFee::try_from(result)?;
 
         Ok(onchain_tx_fee)
     }
+
+    async fn traced_gql_request<Q: GraphQLQuery, U: reqwest::IntoUrl>(
```
bodymindarts, line 333: if you make this an instance method you could avoid the duplication of passing in `self.client` and `self.config.api` everywhere - not a big deal.

bodymindarts (approved):

PR #154, opened 2022-10-10
docs: stablesats developer documentation

enigbe:
## What this PR does
- Adds developer documentation

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -1,0 +1,18 @@
+## Dependencies
+Last tested with the following tools:
+```
+$ rustc --version
+rustc 1.63.0 (4b91a6ea7 2022-08-08)
+$ cargo --version
+cargo 1.63.0 (fd9c4297c 2022-07-01)
+$ docker --version
+Docker version 20.10.18, build b40c2f6
+$ direnv --version
+2.32.1
+```
+
+## How to get started
+
+1. Clone the [galoy]() backend and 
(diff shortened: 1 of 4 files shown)


DEV.md
```diff
@@ -0,0 +1,143 @@
+# `stablesats`
+- [Dependencies](#dependencies)
+    - [Tools](#tools)
+- [Getting started](#getting-started)
+    - [Local Development Mode](#local-development-mode)
+    - [Staging Development Mode](#staging-development-mode)
+- [How to run stablesats](#how-to-run-stablesats)
```
bodymindarts, line 7: information about staging should now be public.

Makefile
```diff
@@ -30,8 +30,13 @@ clean-deps:
 start-deps:
 	docker compose up -d integration-deps
 
+start-deps-local:
```
bodymindarts, line 33: Don't understand why we need this and what it has to do with docs.

enigbe: Running stablesats with a local galoy backend. We don't need to have redis and otel running since we already have them provisioned with galoy.

DEV.md
enigbe: I will update the section on staging.

enigbe: Also removed the section on staging as discussed

bodymindarts (approved):

PR #157, opened 2022-10-12
Fix okex client balance data fields

sebastienverreault:
Fetching the proper balance fields and extending to get free, used and total

### DIFF
--- a/okex-client/src/client/mod.rs
+++ b/okex-client/src/client/mod.rs
@@ -191,10 +191,11 @@ impl OkexClient {
             .await?;
 
         let funding_balance = Self::extract_response_data::<FundingBalanceData>(response).await?;
-        let balance = Decimal::from_str_exact(&funding_balance.avail_bal)?;
 
         Ok(AvailableBalance {
-            amt_in_btc: balance,
+            free_amt_in_btc: funding_balance.avail_bal,
+            used_amt_in_btc: funding_balance.frozen_bal,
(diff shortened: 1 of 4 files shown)


bodymindarts (approved):

okex-client/src/client/mod.rs
```diff
@@ -214,7 +215,9 @@ impl OkexClient {
         let trading_balance = Self::extract_response_data::<TradingBalanceData>(response).await?;
 
         Ok(AvailableBalance {
-            amt_in_btc: trading_balance.details[0].cash_bal,
+            free_amt_in_btc: trading_balance.details[0].avail_eq,
```
enigbe, line 218: In the event that the exchange returns an empty `trading_balance.details` array, it is helpful to check beforehand and to handle the error if that is the case. Something similar to this:
```
let trading_balance_details = trading_balance
            .details
            .first()
            .ok_or(OkexClientError::EmptyTradingBalanceData)?;

Ok(AvailableBalance {
    free_amt_in_btc: trading_balance_details.avail_eq,
    used_amt_in_btc: trading_balance_details.frozen_bal,
    total_amt_in_btc: trading_balance_details.eq,
})
```

enigbe (approved):

okex-client/src/client/mod.rs
bodymindarts: Yeah good point

PR #161, opened 2022-10-14
Feat funding

sebastienverreault:
Missing the deposit/withdrawals aka transfers table in pg for persistence
Missing funding & trading account fund transfer management
A few other TODOs
and some actual testing (units might not align for ex.)


bodymindarts:
Boilerplate looks okay so far but from scanning the code I think funding should not actually be its own process. We should embed it into the hedging process.

You are sharing the same database which is a big no-go and many of the triggers you are listening to are also the same. Finally we will probably never want to deploy just 1 without the other so all in all I think it’s just another type of action / job rather than its own process

sebastienverreault:
Yeah the SQL was not supposed to be committed, it's the hedging code as-is, I had it there as reference.
I'm adding a new table in the same database.
And will merge the funding module to the hedging next.

bodymindarts (commented):
I think there may have been a misunderstanding. We need more granular `RebalanceActions`...

other than that it looks okay - you are eating some critical errors that at least need to be reported and you've added some `clippy` exceptions in tests that I don't understand.

hedging/src/job/adjust_funding.rs
```diff
@@ -0,0 +1,222 @@
...
+    let fees = okex.get_onchain_fees().await?;
+    span.record("onchain_fees", &tracing::field::display(&fees));
+
+    let action = determine_action(
+        target_liability_in_cents,
+        current_position.usd_cents.into(),
+        trading_available_balance.total_amt_in_btc,
+        last_price_in_usd_cents,
+        funding_available_balance.total_amt_in_btc,
+        is_internal_transfer_pending,
```
bodymindarts, line 87: I don't think `determine_action` should depend on this. It should get the current state and determine what to do. Anything inflight is irrelevant. If there is something in flight the database won't reserve a 'slot' and no action will be performed. Hence we already have an implicit `DoNothing` whenever something is in flight.

The only thing that would make sense here is to input the `amount` that is inflight. Then we could figure out if additional action should be taken. But that makes the logic more complicated. The idea of the `slot` reservation is to simplify things and guarantee that only 1 thing is happening at a time and nothing will get kicked off if something is already in flight.

The logic in `determin_action` should become simpler without taking in flight into account.

hedging/src/job/adjust_funding.rs
```diff
@@ -0,0 +1,222 @@
...
+        target_liability, current_position, last_price_in_usd_cents, funding_available_balance,
+        trading_available_balance, onchain_fees, action, external_client_transfer_id, internal_client_transfer_id,
+        transferred_funding) err)]
+pub(super) async fn execute(
+    correlation_id: CorrelationId,
+    synth_usd_liability: SynthUsdLiability,
+    okex: OkexClient,
+    okex_transfers: OkexTransfers,
+    galoy: GaloyClient,
+    hedging_funding_config: HedgingFundingConfig,
```
bodymindarts, line 25: I think instead of passing config in here we should pass `AdjustFunding` or `AdjustHedging` structs around. I will explain more below.

hedging/src/job/mod.rs
```diff
@@ -9,33 +10,48 @@ use uuid::{uuid, Uuid};
...
+    FundingSectionConfig, HedgingSectionConfig,
+};
 
 const POLL_OKEX_ID: Uuid = uuid!("00000000-0000-0000-0000-000000000001");
 
 #[derive(Debug, Clone)]
 struct OkexPollDelay(std::time::Duration);
 
+#[derive(Debug, Clone)]
+pub struct HedgingFundingConfig(pub HedgingSectionConfig, pub FundingSectionConfig);
```
bodymindarts, line 31: Instead of passing the config lets pass the (currently missing) structs.

hedging/migrations/20221013061329_add-okex-transfers.up.sql
```diff
@@ -0,0 +1,31 @@
+-- Add up migration script here
```
bodymindarts, line 1: Commented code.

hedging/migrations/20221013061329_add-okex-transfers.up.sql
```diff
@@ -0,0 +1,31 @@
...
+  transfer_to VARCHAR(128) NULL,
+
+  target_usd_exposure NUMERIC NOT NULL,
+  current_usd_exposure NUMERIC NOT NULL,
+  trading_btc_used_balance NUMERIC NOT NULL,
+  trading_btc_total_balance NUMERIC NOT NULL,
+  current_usd_btc_price NUMERIC NOT NULL,
+  funding_btc_total_balance NUMERIC NOT NULL,
+
+  -- complete BOOLEAN NOT NULL GENERATED ALWAYS AS (CASE WHEN state = 'pending' THEN FALSE ELSE TRUE END) STORED,
```
bodymindarts, line 24: Commented code.

hedging/src/adjustment_action.rs
```diff
@@ -63,15 +59,18 @@ impl AdjustmentAction {
 pub fn determine_action(
     abs_liability: SyntheticCentLiability,
     signed_exposure: SyntheticCentExposure,
+    HedgingFundingConfig(hedging_config, _): HedgingFundingConfig,
```
bodymindarts, line 62: The difference between config and the other args is its lifecycle. Config doesn't change at runtime - the other arguments do.

This means we want to instantiate a (long living) struct that can keep the config internally:

```
pub struct HedgingAdjustment {
  pub config: HedgingConfig
}

impl HedgingAdjustment {
  pub fn determin_action(...) -> AdjustmentAction
}
```

The same applies to the funding config. Then we don't need to wire the config through all the layers - just the instantiated struct.

hedging/src/okex_transfers/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+                trading_btc_used_balance,
+                trading_btc_total_balance,
+                current_usd_btc_price,
+                funding_btc_total_balance,
+                state
+            ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16)"#,
+            String::from(id.clone()),
+            Uuid::from(reservation.shared.correlation_id),
+            reservation.shared.action_type,
+            reservation.transfer_type,
```
bodymindarts, line 81: I'm not sure we should have 2 slots for the 2 different transfer types. Just having 1 slot will simplify the logic.

hedging/src/rebalance_action.rs
```diff
@@ -0,0 +1,763 @@
...
+    total_collateral_in_btc: Decimal,
+    btc_price_in_cents: Decimal,
+    funding_btc_total_balance: Decimal,
+    internal_transfer_pending: bool,
+    external_transfer_pending: bool,
+    HedgingFundingConfig(hedging_config, funding_config): HedgingFundingConfig,
+) -> RebalanceAction {
+    let abs_liability_in_btc = abs_liability_in_cents / btc_price_in_cents;
+    let abs_exposure_in_btc = Decimal::from(signed_exposure_in_cents).abs() / btc_price_in_cents;
+    if abs_exposure_in_btc.is_zero()
```
bodymindarts, line 91: I think this whole block of `if .. else if ... else if` etc. can probably be cleaned up via 1 level of nesting or a `match` clause or perhaps sub functions that describe what the branch signifies. As is its not possible for me to understand what condition is actually being checked here.

hedging/src/adjustment_action.rs
sebastienverreault: great thx for that

hedging/src/rebalance_action.rs
sebastienverreault: I have factored out the second level into sub functions finally, 
each first level conditions have a specific  meaning that would be lost otherwise. 
But each branch is either a deposit or a withdrawal action (and "internal"/"external" depends on the funding available)

hedging/migrations/20221013061329_add-okex-transfers.up.sql
```diff
@@ -3,10 +3,9 @@ CREATE TABLE okex_transfers (
   client_transfer_id VARCHAR(32) PRIMARY KEY,
   correlation_id UUID NOT NULL,
   
-  action VARCHAR(20) NOT NULL CHECK (action in ('withdraw-all', 'withdraw', 'deposit')),
-  transfer_type VARCHAR(20) NOT NULL CHECK (transfer_type in ('internal', 'external')),
+  action VARCHAR(32) NOT NULL CHECK (action in ('transfer-trading-to-funding', 'transfer-funding-to-trading', 'deposit', 'withdraw')),
 
-  currency VARCHAR(20) NOT NULL,
+  currency VARCHAR(16) NOT NULL,
```
bodymindarts, line 8: I wouldn't use `VARCHAR(x)`... if it turns our the assumption of `x` is wrong its tricky to change. There are no performance benefits over `VARCHAR`.

bodymindarts (approved):

PR #172, opened 2022-10-27
Feat add get onchain fees

sebastienverreault:


okex-client/src/client/okex_response.rs
```diff
@@ -21,6 +21,17 @@ pub struct DepositAddressData {
     pub selected: bool,
 }
 
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct OnchainFeesData {
+    pub ccy: String,
+    pub chain: String,
```
bodymindarts, line 28: Do we need to expose `chain` and `ccy`? I can't think of a use case where this would be relevant to the caller.

enigbe (approved):

PR #173, opened 2022-11-01
feat: Kollider-integration

ngutech21:
This is the first step of adding support for the Kollider-exchange to stablesats: 

Adds a new crate that receives price-ticks from the Kollider-API and pushes them on the Redis pubsub.

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -30,6 +30,8 @@ pub struct Config {
     pub okex: OkexClientConfig,
     #[serde(default)]
     pub hedging: HedgingConfigWrapper,
+    #[serde(default)]
+    pub kollider_price_feed: KolliderPriceFeedConfigWrapper,
 }
 
 pub struct EnvOverride {
@@ -102,6 +104,17 @@ impl Default for PriceFeedConfigWrapper {
             enabled: true,
             config: PriceFeedConfig::default(),
         }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct KolliderPriceFeedConfigWrapper {
+    #[serde(default = "bool_true")]
+    pub enabled: bool,
+}
+impl Default for KolliderPriceFeedConfigWrapper {
+    fn default() -> Self {
+        Self { enabled: true }
     }
 }
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -111,10 +111,16 @@ impl Default for PriceFeedConfigWrapper {
 pub struct KolliderPriceFeedConfigWrapper {
     #[serde(default = "bool_true")]
     pub enabled: bool,
+
+    #[serde(default)]
+    pub config: KolliderPriceFeedConfig,
 }
 impl Default for KolliderPriceFeedConfigWrapper {
     fn default() -> Self {
-        Self { enabled: true }
+        Self {
+            enabled: true,
+            config: PriceFeedConfig::default,
+        }
     }
 }
 
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,4 +1,5 @@
 use anyhow::Context;
+use kollider_price::config::KolliderPriceFeedConfig;
 use serde::{Deserialize, Serialize};
 use std::path::Path;
 
@@ -119,7 +120,7 @@ impl Default for KolliderPriceFeedConfigWrapper {
     fn default() -> Self {
         Self {
             enabled: true,
-            config: PriceFeedConfig::default,
+            config: KolliderPriceFeedConfig::default(),
         }
     }
 }
--- a/kollider-price/src/convert.rs
+++ b/kollider-price/src/convert.rs
@@ -1,0 +1,23 @@
+use super::price_feed::error::KolliderPriceFeedError;
+use super::price_feed::KolliderPriceTicker;
+
+use shared::{
+    payload::{
+        ExchangeIdRaw, InstrumentIdRaw, KolliderBtcUsdSwapPricePayload, PriceMessagePayload,
+        PriceRatioRaw,
+    },
+    time::TimeStamp,
+};
+
+impl TryFrom<KolliderPriceTicker> for KolliderBtcUsdSwapPricePayload {
+    type Error = KolliderPriceFeedError;
+    fn try_from(value: KolliderPriceTicker) -> Result<Self, Self::Error> {
+        Ok(KolliderBtcUsdSwapPricePayload(PriceMessagePayload {
+            exchange: ExchangeIdRaw::from("Kollider"), // FIXME
+            instrument_id: InstrumentIdRaw::from(value.symbol),
+            timestamp: TimeStamp::now(), //FIXME
+            ask_price: PriceRatioRaw::from_one_btc_in_usd_price(value.best_ask),
+            bid_price: PriceRatioRaw::from_one_btc_in_usd_price(value.best_bid),
+        }))
+    }
+}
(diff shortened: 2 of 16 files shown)


kollider-price/src/convert.rs
```diff
@@ -0,0 +1,23 @@
...
+        PriceRatioRaw,
+    },
+    time::TimeStamp,
+};
+
+impl TryFrom<KolliderPriceTicker> for KolliderBtcUsdSwapPricePayload {
+    type Error = KolliderPriceFeedError;
+    fn try_from(value: KolliderPriceTicker) -> Result<Self, Self::Error> {
+        Ok(KolliderBtcUsdSwapPricePayload(PriceMessagePayload {
+            exchange: ExchangeIdRaw::from("Kollider"), // FIXME
```
enigbe, line 16: Akin to what is done [here](https://github.com/GaloyMoney/stablesats-rs/blob/1edc01cc9beaec918a1d432c135701e28acd1c7d/shared/src/payload/constants.rs#L1), you can create a constant for the exchange ID.

It appears that Kollider has its unique instrument ID, i.e. `BTCUSD.PERP` but do not return a timestamp with the response from the tickers channel. Unless @bodymindarts has a different opinion, creating a timestamp on the fly as the ticks are received is fine

kollider-price/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    use url::Url;
+
+    #[tokio::test]
+    async fn test_get_price() -> anyhow::Result<()> {
+        let config = KolliderPriceFeedConfig {
+            url: Url::parse("wss://testnet.kollider.xyz/v1/ws/")?,
+        };
+        let mut stream = subscribe_price_feed(config).await?;
+        if let Some(tick) = stream.next().await {
+            println!("first tick connect: {:?}", tick);
```
enigbe, line 44: Given that you are testing here, maybe make some assertion about the data received. This will always pass.

kollider-price/src/price_feed/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    let subscribe_args = serde_json::json!({
+        "type": "subscribe",
+        "symbols": ["BTCUSD.PERP"],
+        "channels": ["ticker"]
+    })
+    .to_string();
+    let item = Message::Text(subscribe_args);
+
+    sender.send(item).await.unwrap();
```
enigbe, line 28: With `.unwrap()` the program will panic if there's an error. We generally avoid using it in production code, preferring to propagate errors.

```
sender.send(item).await?;
```

kollider-price/src/price_feed/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    Ok(Box::pin(receiver.filter_map(|msg| async {
+        let msg = msg.unwrap();
```
enigbe, line 31: The sentiment from [here](https://github.com/GaloyMoney/stablesats-rs/pull/173/files#r1012056502) applies. You can handle them like this too:
```rust
if let Ok(msg) = msg {
    if let Message::Text(txt) = msg {
        if let Ok(tick) = serde_json::from_str::<KolliderPriceTickerRoot>(&txt) {
            return Some(tick.data);
        }
    }
}
```

kollider-price/src/price_feed/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+        if let Message::Text(txt) = msg {
+            println!("tick raw: {}", txt);
```
enigbe, line 33: You can get rid of the print statements when you are done.

cli/src/config.rs
```diff
@@ -105,6 +108,23 @@ impl Default for PriceFeedConfigWrapper {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct KolliderPriceFeedConfigWrapper {
+    #[serde(default = "bool_true")]
```
bodymindarts, line 113: Think this should probably be false by default. `#[serde(default)]` should work as the default for `bool` is `false`.

cli/src/config.rs
```diff
@@ -105,6 +108,23 @@ impl Default for PriceFeedConfigWrapper {
...
+    pub enabled: bool,
+
+    #[serde(default)]
+    pub config: KolliderPriceFeedConfig,
+}
+impl Default for KolliderPriceFeedConfigWrapper {
+    fn default() -> Self {
+        Self {
+            enabled: true,
```
bodymindarts, line 122: Should probably be `false` by default.

kollider-price/src/convert.rs
bodymindarts, line 16: Please remove the `FIXME` comments before merge. Constants shouldn't be defined here.

kollider-price/src/convert.rs
```diff
@@ -0,0 +1,23 @@
...
+            instrument_id: InstrumentIdRaw::from("BTC-USD-SWAP"), // FIXME "BTC-USD-SWAP"
+            timestamp: TimeStamp::now(),               //FIXME
```
bodymindarts, line 18: This is not great but probably fine if kollider doesn't publish a timestamp (it should!).
More accurate would be to create the timestamp in the code that received the message and then calls `KolliderBtcUsdSwapPricePayload::try_from((tick, timestamp))`

kollider-price/src/price_feed/error.rs
```diff
@@ -0,0 +1,17 @@
+use serde_json::Error as SerdeError;
+use shared::pubsub::PublisherError;
+use thiserror::Error;
+
+use tokio_tungstenite::tungstenite::error::Error as TungsteniteError;
+
+#[derive(Error, Debug)]
+pub enum KolliderPriceFeedError {
+    #[error("PriceFeedError - SerdeError: {0}")]
```
bodymindarts, line 9:
```suggestion
    #[error("KolliderPriceFeedError - SerdeError: {0}")]
```

kollider-price/src/price_feed/error.rs
```diff
@@ -0,0 +1,17 @@
...
+    SerializationError(#[from] SerdeError),
+
+    #[error("PriceFeedError - PublisherError: {0}")]
+    PublisherError(#[from] PublisherError),
```
bodymindarts, line 13: See above

kollider-price/src/price_feed/error.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    #[error("PriceFeedError - TungsteniteError: {0}")]
```
bodymindarts, line 15: See above

stablesats.yml
```diff
@@ -38,6 +38,11 @@
   # config:
   #   url: "wss://ws.okx.com:8443/ws/v5/public"
 
+  # kollider_price_feed:
+  # enabled: true
```
bodymindarts, line 42: default should be false

kollider-price/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    use crate::{config::KolliderPriceFeedConfig, price_feed::subscribe_price_feed};
+    use futures::StreamExt;
+    use url::Url;
+
+    #[tokio::test]
+    async fn test_get_price() -> anyhow::Result<()> {
+        let config = KolliderPriceFeedConfig {
+            url: Url::parse("wss://testnet.kollider.xyz/v1/ws/")?,
+        };
+        let mut stream = subscribe_price_feed(config).await?;
```
bodymindarts, line 42: this is not a unit test and thus should IMO be under `src` imo. Given that you already have an e2e test under `kollider-price/tests` and there are no assertions here this test seems redundant.

bodymindarts (changes requested):
It looks pretty good overall!
I think there is a bit of cleanup required pre-merge.
- no `unwrap` in production code
- add tracing / remove `println`

If you have never used honeycomb I'd be happy to jump on a call and give you an intro to observability. TLDR we do not rely on log output but do all debugging via OpenTelemetry.

cli/src/config.rs
ngutech21: I changed the default to false

kollider-price/src/convert.rs
ngutech21: removed the FIXME and defined a constant

kollider-price/src/price_feed/mod.rs
ngutech21: done

PR #174, opened 2022-11-02
fix: use 'delayed' fee for get_cents_from_sats_for_future_sell

vindard:
## Description

Fixes a discrepancy first discovered when doing tests for `galoy` application code https://github.com/GaloyMoney/galoy/pull/1754#discussion_r991368855


bodymindarts (approved):
Thanks! Good catch :-)

PR #177, opened 2022-11-07
Make price-server configurable to use Kollider instead of Okex

ngutech21:
* Adds a config app: use_okex to the stablesats.yml to configure which exchange to use for receiving prices
* If no config is provided the default Okex is used


ngutech21:
This PR has been replaced by a more sophisticated solution: https://github.com/GaloyMoney/stablesats-rs/pull/212

PR #178, opened 2022-11-07
fix: implement ref-to-ref conversion for order book

enigbe:
## What this PR does?
- borrows a reference of complete order book instead of returning a clone

### DIFF
--- a/okex-price/src/order_book/book.rs
+++ b/okex-price/src/order_book/book.rs
@@ -204,8 +204,8 @@ impl OrderBookCache {
         Ok(())
     }
 
-    pub fn latest(&self) -> CompleteOrderBook {
-        self.current.clone()
+    pub fn latest(&self) -> &CompleteOrderBook {
+        &self.current
     }
 }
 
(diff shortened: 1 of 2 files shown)


okex-price/src/order_book/book.rs
```diff
@@ -194,6 +195,12 @@ pub struct OrderBookCache {
     current: CompleteOrderBook,
 }
 
+impl AsRef<CompleteOrderBook> for CompleteOrderBook {
```
bodymindarts, line 198: This should be redundant

okex-price/src/order_book/book.rs
```diff
@@ -204,8 +211,8 @@ impl OrderBookCache {
         Ok(())
     }
 
-    pub fn latest(&self) -> CompleteOrderBook {
-        self.current.clone()
+    pub fn latest(&self) -> &CompleteOrderBook {
+        self.current.as_ref()
```
bodymindarts, line 215: I'm really surprised. Doesn't `&self.current` work?

enigbe: I am uncertain why it didn't work before but I updated my IDE and this works now. Also removed the implementation of `AsRef`

bodymindarts (approved):

PR #196, opened 2022-11-21
revert: weighted price from order book cache

enigbe:
Reverts GaloyMoney/stablesats-rs#181


bodymindarts:
The correct semantic commits prefix for reverts is `revert:`

PR #214, opened 2022-12-01
Use multiple exchanges as source to provide prices

ngutech21:
Adds a new section to stablesats.yml to configure which exchanges to use for providing prices. 
* Each exchange-config contains a weight that will be used to compute the prices.
* The okex-config has moved to the exchanges section in stablesats.yml. So there is one single place to configure all exchange related settings.
* At least one exchange has to be configured, otherwise the cli won't start


cli/src/app.rs
```diff
@@ -174,17 +186,22 @@ async fn run_cmd(
...
         let (snd, recv) = futures::channel::mpsc::unbounded();
         checkers.insert("hedging", snd);
-        handles.push(tokio::spawn(async move {
-            let _ = hedging_send.try_send(
-                hedging::run(recv, hedging.config, okex, pubsub)
-                    .await
-                    .context("Hedging error"),
-            );
-        }));
+        let exchanges = exchanges.clone();
```
bodymindarts, line 194: If you destructure `exchanges` you don't have to clone the whole struct:
```
let ExchangesConfig { okex: okex_cfg, kollider: kollider_cfg } = exchanges;
```
then later
```
if let Some(okex_cfg) = okex_cfg
```
or something like that.

price-server/src/app/mod.rs
```diff
@@ -1,43 +1,137 @@
...
         pubsub_cfg: PubSubConfig,
+        exchanges_cfg: ExchangeConfigAll,
     ) -> Result<Self, PriceAppError> {
+        let health_check_trigger = Arc::new(RwLock::new(health_check_trigger));
+
+        let mut providers: HashMap<String, (Box<dyn PriceProvider + Sync + Send>, Decimal)> =
+            HashMap::new();
+
+        if let Some(config) = exchanges_cfg.kollider.as_ref() {
+            let kollider_price_cache = ExchangeTickCache::new(Duration::seconds(30));
```
bodymindarts, line 46: should exctract the `cache` time into a constant or a config setting. Using the same magic number here and line 60 is not good.

price-server/src/app/mod.rs
```diff
@@ -1,43 +1,137 @@
...
+                .await;
+            }
+        });
+
+        Ok(())
+    }
+
+    async fn subscribe_kollider(
+        pubsub_cfg: PubSubConfig,
+        health_check_trigger: Arc<RwLock<HealthCheckTrigger>>,
```
bodymindarts, line 121: I don't think passing the `health_check_trigger` around behind a `RwLock` will work like this. This way _either_ this _or_ the other method will get a handle to the health checker... but we want both subscribers to be checked on each health_check... so you must spawn off the health check with clones of the subscribers or something like that.

price-server/src/app/mod.rs
```diff
@@ -77,12 +174,11 @@ impl PriceApp {
...
     ) -> Result<UsdCents, PriceAppError> {
-        let cents = self
-            .price_cache
-            .latest_tick()
-            .await?
-            .sell_usd()
-            .cents_from_sats(sats);
+        let cents = UsdCents::from_decimal(
+            self.price_mixer
+                .apply(|p| *p.sell_usd().cents_from_sats(sats.clone()).amount())
```
bodymindarts, line 179: I'm surprised `sats.clone()` is requires. `Decimal` implemented copy.

price-server/src/price_mixer.rs
```diff
@@ -0,0 +1,134 @@
...
+                    prev_error = Some(err);
+                    continue;
+                }
+            };
+            total_weights += weight;
+            total += f(&side_picker) * weight;
+        }
+
+        if let Some(prev_error) = prev_error {
+            Err(prev_error)
```
bodymindarts, line 53: If you return an error here there is no point in having a `continue` at line 45. You may as well early return. Actually we want to return `Ok` even on partial error.

price-server/src/app/error.rs
```diff
@@ -12,4 +12,7 @@ pub enum PriceAppError {
     SubscriberError(#[from] SubscriberError),
     #[error("PriceAppError - ExchangePriceCacheError: {0}")]
     ExchangePriceCacheError(#[from] ExchangePriceCacheError),
+
+    #[error("PriceAppError - DecimalError: {0}")]
+    DecimalError(#[from] rust_decimal::Error),
```
bodymindarts, line 17: I'm a bit surprised to see this error included. Where can it originate from? Can't we rule it out?

bodymindarts (commented):
It looks good overall...

I think there are just 2 significant issues left
- we shouldn't return an error as long as any of the exchanges is online
- health check design is wrong... each tick needs to check each subscriber.

Other than that its just nits that are easy to fix.

price-server/src/price_mixer.rs
ngutech21: I saved the previous error to make the first integration test from the price-server work: This test checks if the error NoPriceAvailabe gets propagated, if there is no tick available. 
Your comment makes sense. I just wanted to point out, that this will change existing behavior

bodymindarts: If none of the exchanges are available we should return an error.

price-server/src/app/mod.rs
bodymindarts: It should be easy to add [here](https://github.com/GaloyMoney/stablesats-rs/blob/main/price-server/src/currency/mod.rs#L18)

price-server/src/app/error.rs
ngutech21: It originated from get_cents_per_sat_exchange_mid_rate()
https://github.com/GaloyMoney/stablesats-rs/blob/d989a8715b60c12e241ff937057aca4eeb682855/price-server/src/app/mod.rs#L266

I fixed it by converting the decimal to sats before calling try_from.

price-server/src/app/mod.rs
ngutech21: I can not implement Copy for the Currency-struct because the inner struct Money does not implement Copy. Any ideas on how to fix this?

PR #289, opened 2023-02-15
feat: exchange allocation

sandipndev:
- [x] Ledger template
- [ ] Fund the accounts at hedge
- [x] Test


ledger/src/templates/exchange_allocation.rs
```diff
@@ -0,0 +1,106 @@
...
+        let tx_input = TxInput::builder()
+            .journal_id(format!("uuid('{STABLESATS_JOURNAL_ID}')"))
+            .effective("params.effective")
+            .metadata("params.meta")
+            .description("'User buys USD'")
+            .build()
+            .expect("Couldn't build TxInput");
+        let entries = vec![
+            EntryInput::builder()
+                .entry_type("'EXCHANGE_ALLOCATION_OMNIBUS_DR'")
```
bodymindarts, line 73: The template doesn't touch an omnibus. You don't need this.
The transfer is from `STABLESATS_LIABILITY_ID: Uuid = uuid!("20000000-2100-0000-0000-000000000000");` to the `OKEX_ALLOCATION` - we're skipping the spot / hedge distribution (for now) and going straight from liability -> diversified exchange allocation.

sandipndev: Okay, thanks for the context here. Will make the change.

ledger/tests/ledger.rs
```diff
@@ -81,3 +81,30 @@ async fn user_buys_and_sells_usd() -> anyhow::Result<()> {
...
+                meta: ExchangeAllocationMeta {
+                    timestamp: chrono::Utc::now(),
+                },
+            },
+        )
+        .await?;
+
+    // let balance = ledger
+    //     .balances()
+    //     .get_ledger_account_balance(DERIVATIVE_ALLOCATIONS_OKEX_ID, "USD");
```
sandipndev, line 107: @bodymindarts 
`DERIVATIVE_ALLOCATIONS_OKEX_ID` is a `pub(super)` constant. With visibility not into it, should we define a `get_derivative_alocations_balance` in `balance.rs`?

sandipndev:
Questions:
- When we hedge, we can either buy or sell. Should selling be treated as negative derivative allocation?
- How do we account initial derivative allocation onto the ledger?

PR #310, opened 2023-03-11
add recommended postgres port mapping for stablesats dev postgres dbs

Kodylow:
took me a couple minutes to figure out why I couldn't get the db setup to work switching the port mapping to 5433. Works on the first try if you set the port correctly but just reading this you wouldn't know that kratos is already using 5433 so can't do the normal " postgres default port + 1"

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -55,7 +55,7 @@ cd stablesats
 ```
 direnv allow
 ```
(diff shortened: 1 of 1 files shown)


bodymindarts:
Just FYI the docs here to run with a local galoy backend are out of date - we typically are running it against the galoy backend running in staging. Also there is just 1 pg running now - not multiple.

It is theoretically still possible to run against a local backend - but  the docs will require updating more than this to reflect how.

Kodylow:
gotcha, is there a better source of truth for getting the dev env set up? happy to update these docs if you need someone to, just trying to get things building so I can play with it as I pull out and put together module pieces for integrating to fedimint

PR #311, opened 2023-03-14
okex -> okx brand rename

Kodylow:
First thing everyone at the office has said is "why is it okex we're okx now" as I showed stablesats to everyone haha. Just used grep and sed to case sensitive change the Okex -> Okx. Passes local tests I can run like you showed me this morning but I don't have the a bitfinex api key to run those tests (US Person).


bodymindarts:
For the future can you please follow https://www.conventionalcommits.org/en/v1.0.0/

bodymindarts (commented):
I've started reviewing this - but its much too much.

You need to be more surgical. You have mixed code refactoring with breaking changes.
Things that break are:
- local dev workflow
- production environment (charts need updating)
- ci system
- database connectivity
- Especially migrations must be immutable and need a lot of care.

Don't forget we are running this in production with a significant amount of funds at risk. We cannot do blanket changes like this just because you got it working on your machine.

I'm happy to accept PR's that fix the naming but you will have to submit them in 3 or 4 phases taking care of 1 breaking element at a time so we can change the other parts of our automation system that is injecting the values - etc.

Kodylow:
Got it, thanks for the fast review and I'll make sure to follow those going forward. 

I just force pushed with a new conventional commit history just updating the README documentation and only touching the okex-crate's variable and path naming without touching environment variables. I can close this this PR and reopen as a new one to clean it up and will do follow on PRs here 1 crate at a time without touching env variables.

PR #321, opened 2023-03-15
fix: zero exposure from api response

sebastienverreault:


hedging/src/okex/job/poll_okex.rs
```diff
@@ -35,8 +35,7 @@ pub async fn execute(
             Ok(details) => {
                 okex_orders.update_order(details).await?;
             }
-            Err(OkexClientError::OrderDoesNotExist)
-            | Err(OkexClientError::ParameterClientIdNotFound) => {
+            Err(OkexClientError::OrderDoesNotExist) => {
```
bodymindarts, line 38: Why this diff?

PR #372, opened 2023-06-01
feat: galoy_transactions unpaired polling

sebastienverreault:
Adding an extra daily poll which starts with a cursor that spans over the oldest unpaired transaction.


migrations/20230608050451_galoy-transactions-unpaired-last-check-at.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE galoy_transactions ADD COLUMN unpaired_last_checked_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp();
```
bodymindarts, line 1: why not `NOW()` as default?

bodymindarts (approved):

PR #396, opened 2023-07-31
chore(deps): bump fred to 6.3.0

thevaibhav-dixit:


shared/src/pubsub/subscriber.rs
```diff
@@ -66,32 +66,26 @@ impl Subscriber {
...
-                    if channel == <M as MessagePayload>::channel() {
-                        if let RedisValue::String(v) = value {
-                            if let Ok(msg) = serde_json::from_str::<Envelope<M>>(&v) {
-                                if msg.payload_type == <M as MessagePayload>::message_type() {
-                                    return Some(Ok(msg));
-                                }
+        let (mut snd, recv) = channel(100);
+        let mut timestamp_sender = self.timestamp_sender.clone();
+        tokio::spawn(async move {
+            while let Ok(message) = message_stream.recv().await {
```
bodymindarts, line 76: Why change it to a `while` loop?

PR #398, opened 2023-08-04
feat: remove redis for exchange position tracking

bodymindarts:


ledger/src/lib.rs
```diff
@@ -67,6 +72,84 @@ impl Ledger {
...
+        exchange_position_id: uuid::Uuid,
+        exchange_id: String,
+        instrument_id: String,
+    ) -> Result<(), LedgerError> {
+        let current_balance = self
+            .balances()
+            .exchange_position_account_balance(exchange_position_id)
+            .await?
+            .map(|b| b.settled())
+            .unwrap_or(rust_decimal::Decimal::ZERO);
```
thevaibhav-dixit, line 89: should we return an error here? if the option is None or does the unwrap_or pattern is fine?

ledger/src/balances.rs
```diff
@@ -34,20 +34,36 @@ impl<'a> Balances<'a> {
...
+
+    pub async fn exchange_position_account_balance(
+        &self,
+        exchange_position_id: impl Into<LedgerAccountId> + std::fmt::Debug,
+    ) -> Result<Option<AccountBalance>, LedgerError> {
+        self.get_ledger_account_balance(HEDGING_JOURNAL_ID, exchange_position_id, self.usd)
+            .await
+    }
+
+    pub async fn hedge_position_omnibus_account_balance(
```
bodymindarts, line 49: consistent use of `exchange_position`

ledger/src/lib.rs
bodymindarts: This is fine

ledger/src/lib.rs
```diff
@@ -67,6 +72,84 @@ impl Ledger {
...
+        let diff = current_balance * CENTS_PER_USD + usd_cents_amount;
+        if diff < rust_decimal::Decimal::ZERO {
```
bodymindarts, line 91: why not `use rust_decimal::Decimal` at the top?

ledger/tests/ledger.rs
```diff
@@ -81,3 +81,56 @@ async fn user_buys_and_sells_usd() -> anyhow::Result<()> {
...
+    let ledger = Ledger::init(&pool).await?;
+
+    let initial_exchange_balance = ledger
+        .balances()
+        .exchange_position_account_balance(OKEX_POSITION_ID)
+        .await?
+        .map(|b| b.settled())
+        .unwrap_or(Decimal::ZERO);
+
+    ledger
```
bodymindarts, line 97: this should not be exposed. Expose `adjust_okex_position` like in the calling code in `poll_okex`

ledger/tests/ledger.rs
```diff
@@ -81,3 +81,56 @@ async fn user_buys_and_sells_usd() -> anyhow::Result<()> {
...
+        )
+        .await?;
+    let balance_after_second_adjustment = ledger
+        .balances()
+        .exchange_position_account_balance(OKEX_POSITION_ID)
+        .await?
+        .unwrap()
+        .settled();
+    assert_eq!(
+        balance_after_second_adjustment - balance_after_first_adjustment,
```
bodymindarts, line 132: I don't understand this assert. You are executing the same template twice and only asserting the effect of the second application. Wouldn't it be better to apply each template once and assert `balance_after_second_adjustment - initial_balance` - that way both templates must have worked.

PR #411, opened 2023-08-25
feat: add bria-client to stablesats

thevaibhav-dixit:
Adds bria to stablesats


bria-client/src/client/bria_client.rs
```diff
@@ -0,0 +1,113 @@
...
+            tonic::metadata::MetadataValue::try_from(key)
+                .map_err(|_| BriaClientError::CouldNotCreateMetadataValue)?,
+        );
+        Ok(request)
+    }
+
+    pub async fn onchain_address(&mut self) -> Result<OnchainAddress, BriaClientError> {
+        match self.get_address().await {
+            Ok(addr) => Ok(addr),
+            Err(_) => self.new_address().await,
```
bodymindarts, line 51: We shouldn't eat this error - only do new_address if the  error is NotFound.

bria-client/src/client/config.rs
```diff
@@ -0,0 +1,19 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+pub struct BriaClientConfig {
+    #[serde(default = "default_url")]
+    pub url: String,
+    #[serde(default)]
+    pub key: String,
```
bodymindarts, line 8: can we be more specific here? `profile_api_key` perhaps?

bria-client/src/client/config.rs
```diff
@@ -0,0 +1,19 @@
...
+    #[serde(default)]
+    pub wallet_name: String,
+    #[serde(default)]
+    pub payout_queue_name: String,
+    #[serde(default)]
+    pub external_id: String,
```
bodymindarts, line 14: How about `onchain_address_external_id`?

bria-client/src/client/bria_client.rs
```diff
@@ -0,0 +1,113 @@
...
+    proto_client: ProtoClient,
+}
+
+impl BriaClient {
+    pub async fn connect(config: BriaClientConfig) -> Result<Self, BriaClientError> {
+        let proto_client = ProtoClient::connect(config.url.clone())
+            .await
+            .map_err(|_| BriaClientError::ConnectionError(config.url.clone()))?;
+
+        if config.key.is_empty() {
```
bodymindarts, line 25: This is too late to validate. It should be asserted in cli.

bria-client/src/error.rs
```diff
@@ -0,0 +1,25 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum BriaClientError {
+    #[error("Couldn't connect to bria at url: {0}")]
+    ConnectionError(String),
+    #[error("Bria key cannot be empty")]
+    EmptyKey,
```
bodymindarts, line 8: I don't think we should need this error.

bria-client/src/error.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[error("Couldn't create MetadataValue")]
+    CouldNotCreateMetadataValue,
+    #[error("Couldn't find address for the given external_id")]
+    AddressNotFound,
```
bodymindarts, line 12: When is this returned? This should be handled internally.

cli/src/config.rs
```diff
@@ -62,7 +67,7 @@ impl Config {
         };
 
         config.db.pg_con = stablesats_pg_con;
-
+        config.bria.key = bria_key;
```
bodymindarts, line 70: Here is where to validate that the key exists.

hedging/src/okex/job/adjust_funding.rs
```diff
@@ -149,10 +151,17 @@ pub(super) async fn execute(
                             &tracing::field::display(String::from(client_id)),
                         );
 
-                        let amount_in_sats = amount * SATS_PER_BTC;
-                        let memo: String = format!("deposit of {amount_in_sats} sats to OKX");
-                        let _ = galoy
-                            .send_onchain_payment(deposit_address, amount_in_sats, Some(memo), 1)
+                        let amount_in_sats = (amount * SATS_PER_BTC)
```
bodymindarts, line 154: This conversion should happen in the client. Accept a decimal in the interface.

bria-client/src/error.rs
thevaibhav-dixit: when we call ```onchain_address()```, it first calls ```get_address()```. suppose we don't have an address for a given ```onchain_address_external_id``` we use ```Err(BriaClientError::AddressNotFound) => self.new_address().await``` to call ```new_address()```

bria-client/tests/bria_client.rs
```diff
@@ -0,0 +1,50 @@
+use std::env;
+
+use ::bria_client::*;
+
+fn client_configuration() -> BriaClientConfig {
+    let url = "http://localhost:2742".to_string();
+    let key = env::var("BRIA_KEY").expect("BRIA_KEY not set");
```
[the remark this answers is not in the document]
bodymindarts: https://github.com/GaloyMoney/bria/blob/3a1ec581ec0d2f0e79f4ecd8b85fd7ef161be769/src/dev_constants.rs#L4

bodymindarts (approved):

PR #428, opened 2023-09-12
fix: consistent naming of env variables

thevaibhav-dixit:
Renames ```AUTH_CODE``` with ```GALOY_PHONE_CODE``` everywhere.


.env.example
```diff
@@ -11,9 +11,9 @@ export BITFINEX_API_KEY="Bitfinex trading API key"
 export BITFINEX_SECRET_KEY="Bitfinex trading secret key"
 
 # Galoy Backend
-export GALOY_PHONE_NUMBER="Galoy phone number"
+export PHONE_NUMBER="Galoy phone number" # required for local testing.
```
bodymindarts, line 14: why not `GALOY_PHONE_NUMBER`? It would be consistent with `GALOY_PHONE_CODE`?

ci/values.yml
```diff
@@ -18,8 +18,8 @@ okex_passphrase: ((okex-creds.passphrase))
 okex_secret_key: ((okex-creds.secret_key))
 bitfinex_api_key: ((bitfinex-creds.api_key))
 bitfinex_secret_key: ((bitfinex-creds.secret_key))
-phone_number: ((galoy-creds.phone_number))
-auth_code: ((galoy-creds.auth_code))
+galoy_phone_number: ((galoy-creds.phone_number))
+galoy_phone_code: ((galoy-creds.phone_code))
```
bodymindarts, line 22: Changing this requires changes in org

thevaibhav-dixit: okay i am changing it back to ```galoy-creds.auth_code```

bodymindarts (approved):

PR #430, opened 2023-09-13
feat: quotes-server

bodymindarts:


proto/quotes/quote_service.proto
```diff
@@ -0,0 +1,48 @@
...
+
+service QuoteService {
+  rpc GetQuoteToBuyCents(GetQuoteToBuyCentsRequest) returns (GetQuoteToBuyCentsResponse) {}
+  rpc GetQuoteToSellCents(GetQuoteToSellCentsRequest) returns (GetQuoteToSellCentsResponse) {}
+  rpc AcceptQuote(AcceptQuoteRequest) returns (AcceptQuoteResponse) {}
+}
+
+message GetQuoteToBuyCentsRequest {
+  oneof quote_for {
+    uint64 amount_to_sell_in_sats = 1;
```
nicolasburtey, line 12: one might want to "sell $100 worth of bitcoin", so the amount as an input should be dollar, not sats.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
   "shared",
   "ledger",
   "price-server",
+  "quotes-server",
```
nicolasburtey, line 7: does it make sense to be different than the price server?

proto/quotes/quote_service.proto
bodymindarts: That’s what the next line is for. You can specify either the btc amount to sell or the USD amount to buy (your scenario). See the oneof keyword

nicolasburtey: it's a different use case. yours is `buy with a USD amount`. the one I mention i `sell with a USD amount`

bodymindarts: There is a matrix of 2*2*2 = 8 use cases.

The 3 binary variables are:
buy / sell USD
denominate amount in BTC / USB
execute immediately / execute in future

The api of the current price server has each of the 8 use cases as its own function call.
In this api I'm proposing to reduce the number of exposed endpoint to 2 (buy / sell) and pass the other binaries as function variables (denomination / execution).

nicolasburtey: If as a user, I request to sell $100 of bitcoin, but get credited 99.98 because of some math approximation and because if a mismatch of source of truth, then I would think there are hidden fees in here or that the system doesn't really work well.

If stablesats is going to be the source of truth, we can make those shortcuts because it would worsen the user experience and reduce trust in the system

bodymindarts: So in your example the request would call `GetQuoteToBuyCents(GetQuoteToBuyCentsRequest)` because we are buying USD (ie. selling bitcoin). But we want to denominate the amount in `use` ($100 worth). That means in the `oneof` argument I would specify the `amount_to_buy_in_cents = 100`

```
message GetQuoteToBuyCentsRequest {
  oneof quote_for {
    uint64 amount_to_sell_in_sats = 1;
    uint64 amount_to_buy_in_cents = 2;
  }

  bool immediate_execution = 3;
}
=>
GetQUoteToBuyCentsRequest {
  quote_for: { amount_to_buy_in_cents: 100 }
  immediate_execution: false
}
```

This is the same way it works now with the [`GetSatsFromCentsForFutureBuy`](https://github.com/GaloyMoney/stablesats-rs/blob/469344f1c5faf2909858a35c017bf313f9461e45/proto/price/price_service.proto#L11) call.

The only difference here is that we are passing some arguments instead of separating each method call.

nicolasburtey: ok, I misread the initial interface.

If I want to buy $100 of bitcoin, can I get a quote by passing a dollar amount?

bodymindarts: Yes (see example above)

nicolasburtey: ok.

I think `message GetQuoteToBuyCentsRequest`

would be better worded with`message GetQuoteToBuyDollarRequest` or `message GetQuoteToBuyStablesatsRequest` 

because otherwise it's a bit convoluted that we using cents as both the assets and unit of account. 

This may be why I have been mis interpreting the API (I skimed through it and didnt' try to read it thoroughly)

proto/quotes/quote_service.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+package services.quotes.v1;
+
+service QuoteService {
+  rpc GetQuoteToBuyCents(GetQuoteToBuyCentsRequest) returns (GetQuoteToBuyCentsResponse) {}
+  rpc GetQuoteToSellCents(GetQuoteToSellCentsRequest) returns (GetQuoteToSellCentsResponse) {}
+  rpc AcceptQuote(AcceptQuoteRequest) returns (AcceptQuoteResponse) {}
+}
+
+message GetQuoteToBuyCentsRequest {
```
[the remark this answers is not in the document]
bodymindarts: Premature IMO.

vindard:
Another general comment. I have the problem right now where it's not easy to reason about whether I'm using the right quote type for the right circumstance. I'd have to refer back to [this test case](https://github.com/GaloyMoney/galoy/blob/0d785c182aca5ba228a18d0d86e60b1391da48c4/test/integration/services/stablesats.spec.ts#L103) to see how the different prices behave with spreads, and the implications for using the wrong method are that we open ourselves up to arbitrage attacks.

I'm wondering if there's a way to capture this spread behaviour somehow in the api names.

migrations/20231004130543_stablesats_quotes.up.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TYPE direction_enum AS ENUM ('BuyCents', 'SellCents');
+
+CREATE TABLE stablesats_quote (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    direction direction_enum NOT NULL,
+    immediate_execution BOOLEAN NOT NULL,
+    sat_amount BIGINT NOT NULL,
+    cent_amount BIGINT NOT NULL,
```
bodymindarts, line 8: Do these need to be indexed?

quotes-server/src/quote/entity.rs
```diff
@@ -3,6 +3,7 @@ use derive_builder::Builder;
 use serde::{Deserialize, Serialize};
 
 use crate::currency::*;
+use crate::EntityEvents;
```
bodymindarts, line 6: this shouldn't be exposed in top level. Also `use crate::{..}`

quotes-server/src/app/mod.rs
```diff
@@ -72,13 +72,15 @@ impl QuotesApp {
...
-        let _usd_amount = self
+        let usd_amount = self
             .price_calculator
             .cents_from_sats_for_buy(Satoshis::from(sats), immediate_execution)
             .await?;
         let new_quote = NewQuote::builder()
             .direction(Direction::BuyCents)
             .immediate_execution(immediate_execution)
+            .cent_amount(usd_amount)
+            .sat_amount(Satoshis::from(sats))
```
bodymindarts, line 83: redundant `Satoshis::from`.. just set it to a local var the same you use line 77.

quotes-server/src/price/error.rs
```diff
@@ -2,10 +2,13 @@ use thiserror::Error;
 
 use shared::time::*;
 
+use crate::cache::OrderBookCacheError;
```
bodymindarts, line 5: newline missing?

quotes-server/src/server/convert.rs
```diff
@@ -0,0 +1,61 @@
...
+                .cent_amount
+                .amount()
+                .to_u64()
+                .expect("cent_amount should always parse to u64"),
+            expires_at: quote
+                .expires_at
+                .timestamp()
+                .to_u32()
+                .expect("timestamp should always parse to u32"),
+            executed: false, // hardcoded for now
```
bodymindarts, line 34: this is okay but I would hard code it somewhere else.

```
quote.executed()
```
```
implementation Quote
 fn executed() -> bool { false }
```

proto/quotes/quote_service.proto
```diff
@@ -0,0 +1,48 @@
...
+message GetQuoteToSellUsdRequest {
+  oneof quote_for {
+    uint64 amount_to_buy_in_sats = 1;
+    uint64 amount_to_sell_in_cents = 2;
+  }
+
+  bool immediate_execution = 3;
+}
+
+message GetQuoteToSellUsdResponse {
```
bodymindarts, line 36: I think we should probably nest a `message Quote` inside here:
```
message GetQuoteToSellUsdResponse {
  Quote quote = 1
  }
```

ledger/src/lib.rs
```diff
@@ -220,6 +222,34 @@ impl Ledger {
...
+        id: LedgerTxId,
+        params: QuoteBuyUsdParams,
+    ) -> Result<(), LedgerError> {
+        self.inner
+            .post_transaction_in_tx(tx, id, QUOTE_BUY_USD_CODE, Some(params))
+            .await?;
+        Ok(())
+    }
+
+    // todo: should i construct the params from the corelation_id?
```
bodymindarts, line 238: yes

ledger/src/templates/quote_buy_usd.rs
```diff
@@ -0,0 +1,133 @@
...
+use tracing::instrument;
+
+use crate::{constants::*, error::*};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct QuoteBuyUsdMeta {
+    #[serde(with = "chrono::serde::ts_seconds")]
+    pub timestamp: DateTime<Utc>,
+    pub btc_tx_id: String,
+    pub usd_tx_id: String,
```
bodymindarts, line 14: These should not be required

ledger/src/templates/quote_buy_usd.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct QuoteBuyUsdParams {
```
bodymindarts, line 18: Missing fees.

ledger/src/templates/quote_buy_usd.rs
```diff
@@ -0,0 +1,133 @@
...
+    #[instrument(name = "ledger.quote_buy_usd.init", skip_all)]
+    pub async fn init(ledger: &SqlxLedger) -> Result<(), LedgerError> {
+        let tx_input = TxInput::builder()
+            .journal_id(format!("uuid('{STABLESATS_JOURNAL_ID}')"))
+            .effective("params.effective")
+            .metadata("params.meta")
+            .description("'Quote buy USD'")
+            .build()
+            .expect("Couldn't build TxInput");
+        let entries = vec![
```
bodymindarts, line 80: Missing fees.

ledger/src/templates/revert_quote_buy_usd.rs
```diff
@@ -0,0 +1,133 @@
...
+        params.insert("effective", effective);
+        params
+    }
+}
+pub struct RevertQuoteBuyUsd {}
+
+impl RevertQuoteBuyUsd {
+    #[instrument(name = "ledger.revert_quote_buy_usd.init", skip_all)]
+    pub async fn init(ledger: &SqlxLedger) -> Result<(), LedgerError> {
+        let tx_input = TxInput::builder()
```
bodymindarts, line 73: Missing correlation id

quotes-server/src/price/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+pub use fee::*;
+pub use mixer::*;
+pub use tick_converter::*;
+pub use traits::*;
+
+pub struct PriceCalculator {
+    fee_calculator: FeeCalculator,
+    price_mixer: PriceMixer,
+}
+
```
bodymindarts, line 21: ConversionResult<T> {
  final_quote: T,
  sats:
  cents:
  sat_fees:
  cent_fees:
}

ledger/src/error.rs
```diff
@@ -5,4 +5,12 @@ use thiserror::Error;
 pub enum LedgerError {
     #[error("LedgerError - SqlxLedger: {0}")]
     SqlxLedger(#[from] sqlx_ledger::SqlxLedgerError),
+    #[error("LedgerError - Transaction not found")]
+    TransactionNotFound,
+    #[error("LedgerError - MissingTxMetadata")]
+    MissingTxMetadata,
+    #[error("LedgerError - NotFound: {0}")]
+    ExpectedEntryNotFoundInTx(&'static str),
+    #[error("LedgerError - SerdeJson: {0}")]
```
thevaibhav-dixit, line 14: note: remove this

quotes-server/tests/quotes_app.rs
```diff
@@ -0,0 +1,164 @@
+use rust_decimal_macros::dec;
```
bodymindarts, line 1: quote with immediate execution
- get quote
- accept quote => error (already executed)
quote with delayed execution
- get quote
- accept quote => success
- accept quote => error (already executed)

quotes-server/tests/quotes_app.rs
```diff
@@ -0,0 +1,125 @@
...
+        .await;
+    assert!(quote.is_ok());
+
+    let err = app.accept_quote(quote.unwrap().id).await;
+    assert!(matches!(err, Err(QuotesAppError::QuoteAlreadyAccepted(_))));
+
+    let quote = app
+        .quote_cents_from_sats_for_buy(dec!(100_000_000), false)
+        .await;
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
bodymindarts, line 120: why not 2? And magic number...

ledger/src/balances.rs
```diff
@@ -17,6 +17,16 @@ impl<'a> Balances<'a> {
             .await
     }
 
+    pub async fn quotes_usd_liability(&self) -> Result<Option<AccountBalance>, LedgerError> {
+        self.get_ledger_account_balance(STABLESATS_JOURNAL_ID, QUOTES_LIABILITY_ID, self.usd)
+            .await
+    }
+
+    pub async fn quotes_btc_liability(&self) -> Result<Option<AccountBalance>, LedgerError> {
```
bodymindarts, line 25: btc is not a liability

quotes-server/src/app/config.rs
```diff
@@ -0,0 +1,50 @@
...
+impl Default for QuotesConfig {
+    fn default() -> Self {
+        Self {
+            expiration_interval: default_expiration_interval(),
+        }
+    }
+}
+
+fn default_expiration_interval() -> Duration {
+    const DEFAULT_EXPIRATION_INTERVAL: std::time::Duration = std::time::Duration::from_secs(120);
```
bodymindarts, line 48: redundant const. `default...` fn already identifies unique access point + its not shared outside of this file or even outside this function.

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            .direction(Direction::BuyCents)
+            .immediate_execution(immediate_execution)
+            .cent_amount(usd_amount)
+            .sat_amount(sats)
+            .expires_at(expiry_time)
+            .build()
+            .expect("Could not build quote");
+        let quote = self.quotes.create(new_quote).await?;
+        if immediate_execution {
+            self.accept_quote(quote.id).await?;
```
bodymindarts, line 103: Needs to be in transaction

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let usd_amount = self
+            .price_calculator
+            .cents_from_sats_for_buy(sats.clone(), immediate_execution)
+            .await?;
+        let expiry_time = expiration_time_from_duration(self.config.expiration_interval);
+        let new_quote = NewQuote::builder()
+            .direction(Direction::BuyCents)
+            .immediate_execution(immediate_execution)
+            .cent_amount(usd_amount)
+            .sat_amount(sats)
```
bodymindarts, line 97: Missing cent_spread / sat_spread

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            .direction(Direction::SellCents)
+            .immediate_execution(immediate_execution)
+            .cent_amount(usd_amount)
+            .sat_amount(sats)
+            .expires_at(expiry_time)
+            .build()
+            .expect("Could not build quote");
+        let quote = self.quotes.create(new_quote).await?;
+        if immediate_execution {
+            self.accept_quote(quote.id).await?;
```
bodymindarts, line 130: needs to be in transactoi

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            .direction(Direction::BuyCents)
+            .immediate_execution(immediate_execution)
+            .cent_amount(cents)
+            .sat_amount(sat_amount)
+            .expires_at(expiry_time)
+            .build()
+            .expect("Could not build quote");
+        let quote = self.quotes.create(new_quote).await?;
+        if immediate_execution {
+            self.accept_quote(quote.id).await?;
```
bodymindarts, line 184: accept_quote(id) ->
accept_quote_in_tx(quote, tx)

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+
+        if quote.is_accepted() {
+            return Err(QuotesAppError::QuoteAlreadyAccepted(id.to_string()));
+        }
+
+        if quote.is_expired() {
+            return Err(QuotesAppError::QuoteExpired(id.to_string()));
+        }
+
+        quote.accept();
```
bodymindarts, line 201: accept should fail with expired / already accepted errors.

quotes-server/src/app/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+                        usd_cents_amount: Decimal::from(quote.cent_amount.clone()),
+                        satoshi_amount: Decimal::from(quote.sat_amount.clone()),
+                        meta: BuyUsdQuoteAcceptedMeta {
+                            timestamp: Utc::now(),
+                        },
+                    },
+                )
+                .await?;
+        }
+        self.quotes.update(quote).await?;
```
bodymindarts, line 232: not atomic?

shared/src/macros.rs
```diff
@@ -124,3 +124,58 @@ macro_rules! decimal_wrapper_common {
         }
     };
 }
+
+#[macro_export]
+macro_rules! entity_id {
```
bodymindarts, line 129: redundant

sebastienverreault (approved):
spread calcs look good

thevaibhav-dixit:
closing since most of the work on this has been done via multiple smaller pr's.

PR #446, opened 2023-10-14
chore: auto-resubsribe when no msg arrived

bodymindarts:


okex-price/src/error.rs
```diff
@@ -35,4 +35,8 @@ pub enum PriceFeedError {
     InitialFullLoad,
     #[error("PriceFeedError: CheckSumValidation - Can't validate accuracy of depth data")]
     CheckSumValidation,
+    #[error("PriceFeedError: StreamEnded - Stream ended unexpectedly")]
+    StreamEnded,
+    #[error("PriceFeedError: StreamStalled - Stream ended unexpectedly")]
+    StreamStalled,
```
thevaibhav-dixit, line 41: where is this being used ?

bodymindarts: good catch fixed here: https://github.com/GaloyMoney/stablesats-rs/pull/446/commits/72d8abd269c0077254afd0e55790b4d76eb20662

PR #449, opened 2023-10-23
fix: okx deposit idempotency

bodymindarts:
Deposits to okx get identified in a specific state.
State "1" are not yet available for withdraw and thus should still remain in state 'pending' otherwise multiple deposits will be
initiated.

### DIFF
--- a/okex-client/src/client/mod.rs
+++ b/okex-client/src/client/mod.rs
@@ -257,6 +257,7 @@ impl OkexClient {
         })
     }
 
+    /// https://www.okx.com/docs-v5/en/#funding-account-rest-api-get-balance
     #[instrument(name = "okex_client.funding_account_balance", skip(self), err)]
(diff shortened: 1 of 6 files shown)


okex-client/src/client/mod.rs
```diff
@@ -428,10 +435,12 @@ impl OkexClient {
         });
 
         if let Some(deposit_data) = deposit {
+            tracing::Span::current().record("deposit_found", true);
+            tracing::Span::current().record("okex_deposit_state", &deposit_data.state);
             Ok(DepositStatus {
                 state: match &deposit_data.state[..] {
                     "0" => "pending".to_string(),  // waiting for confirmation
-                    "1" => "success".to_string(),  // deposit credited, cannot withdraw
+                    "1" => "pending".to_string(),  // deposit credited, cannot withdraw
```
bodymindarts, line 443: Here is the root cause

PR #471, opened 2023-11-14
feat: add quote entity 

thevaibhav-dixit:


quotes-server/src/quote/entity.rs
```diff
@@ -0,0 +1,206 @@
...
+pub enum QuoteEvent {
+    Initialized {
+        id: QuoteId,
+        direction: Direction,
+        immediate_execution: bool,
+        sat_amount: Satoshis,
+        cent_amount: UsdCents,
+        expires_at: DateTime<Utc>,
+    },
+    Accepted {},
```
bodymindarts, line 29: Perhaps add a timestamp here wdyt?

bodymindarts (approved):

PR #482, opened 2023-11-27
fix: correct rounding for sats/cents spread

thevaibhav-dixit:


quotes-server/src/price/mod.rs
```diff
@@ -49,7 +49,7 @@ impl PriceCalculator {
             .fee_calculator
             .decrease_by_fee(immediate_execution, cents.clone())
             .floor();
-        let cents_spread = UsdCents::from(cents_after_fee.amount() - cents.amount());
+        let cents_spread = UsdCents::from((cents_after_fee.amount() - cents.amount()).round());
```
bodymindarts, line 52: why is this not `cents_after_fee - cents`? Why the round trip via `from` and `amount()`? Seems quite unintuitive. Also let's be explicit about how you are rounding. Its not clear to me that this will always lead to the correct result.

thevaibhav-dixit: i can change the rounding off logic after the review from @sebastienverreault

sebastienverreault (approved):

bodymindarts:
Need to add the spread to the test assertions. Also its better to call floor/ceil on the amount before adding it to the `xxx_by_fee` then you probably don't need to round the spread.

PR #490, opened 2024-01-03
feat: add trading/funding account balance

thevaibhav-dixit:


bitfinex-client/src/client/mod.rs
```diff
@@ -349,6 +350,66 @@ impl BitfinexClient {
...
+    #[instrument(name = "bitfinex_client.trading_account_balance", skip(self), err)]
+    pub async fn trading_account_balance(
+        &self,
+    ) -> Result<TradingAccountBalance, BitfinexClientError> {
+        let mut body: HashMap<String, String> = HashMap::new();
+        body.insert("type".to_string(), Wallet::EXCHANGE.to_string());
+        if self.config.simulated {
+            body.insert("symbol".to_string(), "fUSD".to_string());
+        } else {
+            body.insert("symbol".to_string(), Currency::USD.to_string());
```
thevaibhav-dixit, line 392: @sebastienverreault what is the right value for symbol. i was not sure so used ```fUSD``` to test the implementation.

sebastienverreault: Aren't "fUSD" and Currency::USD the same?
We probably want the balance of the accounts in Currency::BTC, right?
As USD is the exposure we get by shorting the BTC/USD SWAP instrument based on some BTC posted margin.

sebastienverreault: Also the symbol usually refers to an instrument rather than a currency
so it looks like the endpoint is not the right one to get the balance of the account.

The doc says: "Calculate the balance available for orders/offers"
so sounds like it would give a number of contracts, for ex., that you can buy/sell
rather than the balance of the account in BTC or else.

thevaibhav-dixit: You are right probably... I think https://docs.bitfinex.com/reference/rest-auth-info-margin this is the correct endpoint

PR #491, opened 2024-01-06
chore: add funding/trading account balance, fetch deposit and close pos fn's 

thevaibhav-dixit:


bitfinex-client/src/client/mod.rs
```diff
@@ -194,6 +241,41 @@ impl BitfinexClient {
...
+
+        // get the BTC_USD_SWAP position
+        let mut symbol = Instrument::BtcUsdSwap.to_string();
+        if self.config.simulated {
+            symbol = "tTESTBTC:TESTUSD".to_string();
+        }
+        let btc_usd_position = positions
+            .iter()
+            .find(|p| p.symbol == symbol && p.status == "ACTIVE")
+            .ok_or(BitfinexClientError::NoActivePositionAvailable)?;
```
thevaibhav-dixit, line 256: a bit confused as to what should we return if there is no active position on the exchange? 
@sebastienverreault @bodymindarts

bitfinex-client/src/client/bitfinex_response.rs
```diff
@@ -399,10 +459,17 @@ mod tests {
     #[test]
     fn wallet_details() {
         // let response_text = "[[\"exchange\",\"TESTBTC\",0.01,0,0.01,null,null],[\"exchange\",\"TESTUSD\",100,0,100,null,null],[\"exchange\",\"TESTUSDT\",200,0,200,null,null]]";
-        let response_text = "[[\"exchange\",\"TESTBTC\",0.005,0,0.005,null,null],[\"exchange\",\"TESTUSD\",100,0,100,null,null],[\"exchange\",\"TESTUSDT\",200,0,200,null,null],[\"margin\",\"TESTBTC\",0.005,0,0.005,null,null]]";
+        let response_text = "[[\"exchange\",\"TESTBTC\",1e-8,0,1e-8,null,null],[\"margin\",\"TESTUSD\",10921.63184294,0,0,\"Position #169742974 funding cost\",{\"reason\":\"POS_FUNDING_SETTLE\",\"pos_id\":169742974,\"pos_pair\":\"TESTBTC:TESTUSD\"}],[\"exchange\",\"TESTUSDT\",11000,0,11000,null,null],[\"exchange\",\"TESTUSD\",49999,0,49999,null,null]]";
```
thevaibhav-dixit, line 462: had to change the test vector because trading_details which is an optional JSON was not being handled properly.

PR #493, opened 2024-01-09
chore: remove bitfinex-client and bitfinex-price

thevaibhav-dixit:


price-server/src/exchange_tick_cache.rs
```diff
@@ -14,6 +14,7 @@ pub struct ExchangeTickCache {
     config: ExchangePriceCacheConfig,
 }
 
+#[allow(dead_code)]
```
thevaibhav-dixit, line 17: added this because the methods are used in tests

bodymindarts: Then why not `[cfg(test)]`?

bodymindarts (approved):

PR #494, opened 2024-01-18
chore: add tracing headers for bria

thevaibhav-dixit:


bria-client/src/client.rs
```diff
@@ -41,6 +46,18 @@ impl BriaClient {
...
             tonic::metadata::MetadataValue::try_from(profile_api_key)?,
         );
+
+        let context = tracing::Span::current().context();
+        let propagator = TraceContextPropagator::new();
+        let mut headers = HashMap::new();
+        propagator.inject_context(&context, &mut headers);
+        for (key, value) in headers {
+            request.metadata_mut().insert(
+                tonic::metadata::MetadataKey::from_bytes(key.as_bytes())?,
```
bodymindarts, line 56: why as bytes? `PROFILE_API_KEY_HEADER` is injected as str.

thevaibhav-dixit: it need to be converted into ``` K: IntoMetadataKey<Ascii>``` which is not implemented by String directly

bodymindarts: But it must by implemented by `&str` right or it wouldn't work with the constant

bodymindarts: Did you test this? If it shows up correctly in honeycomb with this code I don't care - but I don't think it will work with the key as bytes.

